PR #117, opened 2022-10-06
Update pipenv requirements generation instructions

MVrachev:
With the new pipenv version there is no more a "--requirements" argument and now instead is "pipenv requirements > requirements.txt"

The same is about "dev" requirements.
Now the new command is:
"pipenv requirements --dev > dev-requirements.txt"


MVrachev:
I closed it accidently.

PR #124, opened 2022-10-13
Update pipenv requirements generation instructions

MVrachev:
With the new pipenv version there is no more a "--requirements" argument and now instead is "pipenv requirements > requirements.txt"

The same is about "dev" requirements.
Now the new command is:
"pipenv requirements --dev > dev-requirements.txt"

Also, add a test for that in `tox.ini`.


tox.ini
```diff
@@ -50,7 +50,18 @@ commands =
...
+    pipenv --version
+    bash -c 'diff requirements.txt <(pipenv requirements)'
+    bash -c 'diff requirements-dev.txt <(pipenv requirements --dev)'
+
 [gh-actions]
 python =
     3.9: py39,pep8,lint
-    3.10: py310,pep8,lint
\ No newline at end of file
+    3.10: py310,pep8,lint
```
kairoaraujo, line 67:
```suggestion
    3.10: py310,pep8,lint,requirements,test,test-doc
```

tox.ini
```diff
@@ -50,7 +50,18 @@ commands =
...
+allowlist_externals =
+    bash
+commands =
+    pipenv --version
+    bash -c 'diff requirements.txt <(pipenv requirements)'
+    bash -c 'diff requirements-dev.txt <(pipenv requirements --dev)'
+
 [gh-actions]
 python =
     3.9: py39,pep8,lint
```
kairoaraujo, line 66:
```suggestion
    3.9: py39,pep8,lint,requirements,test,test-doc
```

kairoaraujo (changes requested):
Sorry, but I notice it now.
Not sure if it can be done in another way. Anyway, if there is, we can raise a new issue for it and apply my suggestions.

MVrachev:
I rebased on main and followed your suggestions.

kairoaraujo (approved):
LGTM

kairoaraujo:
Thank you, @MVrachev. I am merging it.

PR #182, opened 2022-11-16
Fix security issue with Tokens when empty scope

### LINKED ISSUE #175: Bug: user authorized without any scope receives all permissions
### What happened?

While testing locally I authorized myself as a user without selecting any scope.
Then I was able to execute multiple API calls even thought in their documentation it's said they require specific scope permissions.though

### What steps did you take?

1. launch a RSTUF API instance locally
2. authorize as an `admin` with password `secret` without selecting any of the scope options
3. executed successfully: `add targets`, `read config` and `check bootstrap status` commands which all require scope permissions

None of the commands in step 3 should be able to be executed.

### Relevant log output

_No response_

### Code of Conduct

kairoaraujo:
The initial design for the built-in Authentication/Authorization was based on multiple users with scopes.
Users, after authentication, would receive a token with all its defined scopes.

Authentication and Authorization are not core components of RSTUF. We simplified this design to have only one user (admin), and the admin can generate Tokens with specific Scopes.
During this design change, we did not change the user fetching all scopes if the scope is empty/None.
We even want to allow the user to disable the
Authentication/Authorization layer (see
https://github.com/vmware/repository-service-tuf/issues/41) and use their  Authentication/Authorization layer.

- Add scope as required
- Add tests to avoid regression
- Hide password in the swagger try out
- Gives a clearer view of the scopes in the documentation (swagger)

Closes #175

### DIFF
--- a/tests/unit/api/test_token.py
+++ b/tests/unit/api/test_token.py
@@ -174,15 +174,7 @@ class TestPostToken:
         }
         assert mocked_bcrypt.checkpw.calls == [pretend.call(b"test", "test")]
 
-    def test_post_with_empty_scope(self, monkeypatch, test_client):
-        mocked_datetime = pretend.stub(
-            utcnow=pretend.call_recorder(
-                lambda: datetime(2019, 6, 16, 9, 5, 00, 355186)
-            )
-        )
-        monkeypatch.setattr(
-            "repository_service_tuf_api.token.datetime", mocked_datetime
-        )
+    def test_post_with_empty_scope(self, test_client):
         url = "/api/v1/token/"
         token_data = {
             "username": "admin",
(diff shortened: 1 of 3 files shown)


rdimitrov (dismissed):

tests/unit/api/test_token.py
```diff
@@ -173,3 +173,40 @@ class User:
             "detail": {"error": "scope 'write:targets' forbidden"}
         }
         assert mocked_bcrypt.checkpw.calls == [pretend.call(b"test", "test")]
+
+    def test_post_with_empty_scope(self, monkeypatch, test_client):
```
MVrachev, line 177: I don't quite understand how can I debug those tests locally from vscode?
How do you execute it?
Do I have always to run it with `pytest` to get this `monkeypatch` fixture?

MVrachev (commented):
I tried locally logging without selecting a scope and it didn't work.
After login, I also tried generating a new scope by creating a token without a scope and that succeeded. I tried then using it for any request, but it didn't work.

Do you think it will be worth forbidding creating a token with an empty scope as a whole?

kairoaraujo:
I think worth it.
But I think we can use it in another issue as it will handle another endpoint and is not critical as the current bug.
Do you agree?

MVrachev:
> I think worth it. But I think we can use it in another issue as it will handle another endpoint and is not critical as the current bug. Do you agree?

Yes, agreed. If you can create another issue we should be fine.
The last thing I wanted to do is actually run the tests locally and I had problems with it.
So, when I managed and see how they work will approve them.

tests/unit/api/test_token.py
kairoaraujo: Yes, `monkeypatch` is a [Mock](https://docs.pytest.org/en/7.1.x/how-to/monkeypatch.html) fixture built-in in the `pytest`.
Not sure if I got the point here.

To the Unit Test I'm Mocking the backend generating the token, the `datetime`. To be honest, in this specific test case isn't required this Mock, once I don't assert "date time" anywhere.

rdimitrov (approved):

MVrachev (approved):

Issue #187, opened 2022-11-22
Bug: Cannot generate a token with 'read:tasks' scopes.


kairoaraujo:
### What happened?

When running the FT, it was shown that it is impossible to generate a token with the scope `'read:tasks'`.

It is visible that `'read:tasks'` is missing in the code:

https://github.com/vmware/repository-service-tuf-api/blob/7b57e1778ec1608cb90d68a1490264e5ba4b3549/repository_service_tuf_api/token.py#L84-L92

### What steps did you take?

a) You can see it in the functional tests: https://github.com/vmware/repository-service-tuf/actions/runs/3526391298/jobs/5914262984

b) Generate a Token using API
1. Login using admin
2. Generate a token `/api/v1/token/new` adding scope `read:tasks`

c) Generate a token using CLI
1. Login using rstuf
2. Generate a token with scope `read:tasks`

### What behavior did you expect?

Create token with `'read:tasks'` is a must to have.

### Relevant log output

```shell
FAILED tests/functional/tokens/test_generate.py::test_admin_uses_http_api_to_generate_a_token[['write:targets', 'read:settings', 'read:tasks']-380] - AssertionError: {'detail': [{'ctx': {'given': 'read:tasks', 'permitted': ['read:bootstrap', 'read:settings', 'read:targets', 'read:tok...tted: 'read:bootstrap', 'read:settings', 'read:targets', 'read:token', 'write:targets'", 'type': 'value_error.const'}]}

FAILED tests/functional/tokens/test_generate.py::test_admin_uses_rstuf_command_line_interface_to_generate_token[write:targets read:targets read:tasks-380] - AssertionError: (1, '\x1b[31m╭─\x1b[0m\x1b[31m Error \x1b[0m\x1b[31m──────────────────────────────────────────────────────────────────...  \x1b[31m│\x1b[0m
554
  \x1b[31m╰──────────────────────────────────────────────────────────────────────────────╯\x1b[0m
555
  ')
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

MVrachev:
I can work on this one.

kairoaraujo:
Closed by #193
- #193 193

PR #193, opened 2022-11-24
Fix the generation of tokens with specific scopes and disallow token with empty scope

### LINKED ISSUE #185: Bug: Generating a new scope by creating a token without a scope succeeds.
### What happened?

It's impossible to login with 'admin' and gets an `access_token` if `scopes` are not given.
See #175 and #182 

However, if you login with 'admin' with the scope 'write:token' and get the 'access_token' you will be able to generate a token without scopes.

Related to #182 comments
Linked to #175 

### What steps did you take?

1. Login using 'admin' and the password `/api/v1/token` with `'write:token'` in the scopes
2. Get the `access_token`
3. Generate a token using the `access_token` in `/api/v1/token/new` with scope empty

Receive an error that scope is required.

### Relevant log output

_No response_

### Code of Conduct

MVrachev:
Fix a bug where the generation of a token with a scope containing `read:tasks` or `delete:targets` lead to an error.
Also, disallow the generation of a token where the `scopes` field is an empty list.

I added regression tests for both changes.

Close #185 #187 

### DIFF
--- a/tests/unit/api/test_token.py
+++ b/tests/unit/api/test_token.py
@@ -104,6 +104,20 @@ class TestPostToken:
         assert response.status_code == status.HTTP_200_OK
         assert "access_token" in response.json()
 
+    def test_post_new_with_empty_scope(self, test_client, token_headers):
+        url = "/api/v1/token/new/"
+        payload = {
+            "scopes": [],
+            "expires": 1,
+        }
+
+        response = test_client.post(url, headers=token_headers, json=payload)
+        assert response.status_code == status.HTTP_422_UNPROCESSABLE_ENTITY
+        assert (
+            response.json()["detail"][0]["msg"]
+            == "ensure this value has at least 1 items"
+        )
+
     def test_post_without_expires(self, monkeypatch, test_client):
         mocked_datetime = pretend.stub(
             utcnow=pretend.call_recorder(
--- a/tests/unit/api/test_token.py
+++ b/tests/unit/api/test_token.py
@@ -86,38 +86,6 @@ class TestPostToken:
         assert response.status_code == status.HTTP_200_OK
         assert "access_token" in response.json()
 
-    def test_post_new(self, test_client, token_headers):
-        url = "/api/v1/token/new/"
-        payload = {
-            "scopes": [
-                "read:bootstrap",
-                "read:settings",
-                "read:tasks",
-                "read:token",
-                "write:targets",
-                "delete:targets",
-            ],
-            "expires": 1,
-        }
-
-        response = test_client.post(url, headers=token_headers, json=payload)
-        assert response.status_code == status.HTTP_200_OK
-        assert "access_token" in response.json()
-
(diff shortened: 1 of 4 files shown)


tests/unit/api/test_token.py
```diff
@@ -77,12 +77,47 @@ def test_post(self, monkeypatch, test_client):
         token_data = {
             "username": "admin",
             "password": "secret",
-            "scope": "write:targets",
+            "scope": (
+                "read:bootstrap read:settings read:tasks read:token "
+                "write:targets delete:targets"
+            ),
```
MVrachev, line 83: Here I added all allowed scopes to make sure we can create a token containing each of them and such bugs as #187 won't happen for other scopes unnoticed.

tests/unit/api/test_token.py
```diff
@@ -77,12 +77,47 @@ def test_post(self, monkeypatch, test_client):
...
         }
         response = test_client.post(url, data=token_data)
         assert response.status_code == status.HTTP_200_OK
         assert "access_token" in response.json()
 
+    def test_post_new(self, test_client, token_headers):
```
kairoaraujo, line 89: nit: Add both tests after `def test_post_with_missing_scope`

kairoaraujo (changes requested):
I would add two more tests

```python
    def test_post_new_with_scopes_with_empty_string(self, test_client, token_headers):
        url = "/api/v1/token/new/"
        payload = {
            "scopes": [''],
            "expires": 1,
        }

        response = test_client.post(url, headers=token_headers, json=payload)
        assert response.status_code == status.HTTP_422_UNPROCESSABLE_ENTITY
        assert "unexpected value" in response.json()["detail"][0]["msg"]
        
    def test_post_new_with_scopes_with_invalid_scope(self, test_client, token_headers):
        url = "/api/v1/token/new/"
        payload = {
            "scopes": ["invalid"],
            "expires": 1,
        }

        response = test_client.post(url, headers=token_headers, json=payload)
        assert response.status_code == status.HTTP_422_UNPROCESSABLE_ENTITY
        assert "unexpected value" in response.json()["detail"][0]["msg"]

MVrachev:
I added the two tests you suggested.
I decided to move the `post_new` token tests next to the rest of the `post_new` tests.
I am not sure what is the connection between the tests I added and the `test_post_with_missing_scope` test. 
Do you think it's better to move the tests there or leave them where I suggest right now?

kairoaraujo:
> I added the two tests you suggested.
> I decided to move the `post_new` token tests next to the rest of the `post_new` tests.
> I am not sure what is the connection between the tests I added and the `test_post_with_missing_scope` test.
> Do you think it's better to move the tests there or leave them where I suggest right now?

I like the way you did it. All `test_pos_*` and all `test_post_new_*`

kairoaraujo (approved):
LGTM! Ready to merge.

PR #198, opened 2022-12-08
docs/tests: add missing delegation path in payload

MVrachev:
In the example payload.json we use for our swagger.json we define the metadata structure.
There we define that "2.targets.json" delegates all target files within the following paths "*/*" and "*/*/*/*".
I noticed we are missing the case where there is a target file in the root directory or in "*".
That's a case we want to test in our functional tests in the umbrella repository and it makes sense to support such targets.


kairoaraujo (approved):

kairoaraujo (commented):
It needs make-docs to have the new doc built.

MVrachev:
It's ready now @kairoaraujo.

MVrachev:
Wait, @kairoaraujo.
We need to resign the metadata if we change the delegation attributes...

kairoaraujo (approved):

MVrachev:
I updated the example payload.json with the one you can get from running:
```
pytest -vv rstuf-cli/tests/unit/cli/admin/test_ceremony.py::TestCeremonyGroupCLI::test_ceremony_start_default_values
```
That way the `2.targets.json` is resigned, but this means that all keys had to be updated as well.

kairoaraujo (approved):

PR #199, opened 2022-12-12
fix: consistent snapshot field not in the task

kairoaraujo:
The TUF Root Metadata `consistent_snasphot: True` was not sent in the task.
The `1.root.json` in the Storage Backend was not saved with the field.

It doesn't impact the client, unless the client uses this file as the initial `root.json`.

Minor: Fix `README.rst` subtitle format.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -61,7 +61,7 @@ to make sure you stay up-to-date with our repository:
 
 
 Installing project requirements
-==============================
+===============================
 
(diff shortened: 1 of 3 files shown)


rdimitrov (approved):
Looks good 👍

MVrachev (approved):
It's good to see that fixed.
Awesome!

Issue #57, opened 2022-08-16
Add option to add task_id to the target custom fields


kairoaraujo:
Add possibility to the target payload include the `task_id` in the Targets custom field

```json
'payload': {
  'targets': '...',
  'add_task_id_to_custom': bool
}
```

It will have registered to the Metadata the task id responsible for adding the task.

PR #219, opened 2023-01-10
Add an option to add task_id to target custom

MVrachev:
Add an option which can be used for debugging to add the "task_id" of the process that adds the target into the "custom" attributes of all targets.
By default this option is enabled as it will add a small overhead, but can be quite useful for debugging in a bad scenario.

Closes https://github.com/vmware/repository-service-tuf-api/issues/57

### DIFF
--- a/docs/source/guide/api/swagger.json
+++ b/docs/source/guide/api/swagger.json
@@ -444,7 +444,8 @@
             "AddPayload": {
                 "title": "AddPayload",
                 "required": [
-                    "targets"
+                    "targets",
+                    "add_task_id_to_custom"
                 ],
                 "type": "object",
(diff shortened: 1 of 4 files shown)


kairoaraujo (changes requested):

MVrachev:
Ready for another review.

kairoaraujo (changes requested):
Can you please exclude the `docs/source/_static` from this PR.
I know this isn't very pleasant, but I will check how to build those diagrams in another way.

kairoaraujo (approved):

MVrachev:
Can we merge this one or do we want someone else to see it as well?

PR #222, opened 2023-01-11
Remove python3.9 and add python3.11 ci build

MVrachev:
I noticed we run build for python3.9 when we say we support only python3.10 onwards.
That's why it seemed unnecessary that we are testing for python3.9.

On the other hand, python 3.11 was introduced back in October 2022 and we can directly add support for it.

Related to:
- https://github.com/vmware/repository-service-tuf-worker/pull/162
- https://github.com/vmware/repository-service-tuf-cli/pull/172

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-versions: [ "3.9", "3.10", ]
+        python-versions: [ "3.10", "3.11" ]
 
     steps:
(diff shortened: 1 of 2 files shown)


kairoaraujo (changes requested):
Can you please add python 3.11 also to the `tox.ini`


https://github.com/vmware/repository-service-tuf-api/blob/d04eaccfe01abdd175d57b323da6c25ad7251b53/tox.ini#L2

https://github.com/vmware/repository-service-tuf-api/blob/d04eaccfe01abdd175d57b323da6c25ad7251b53/tox.ini#L68-L69

kairoaraujo (approved):

PR #239, opened 2023-01-26
fix: add_task_id_to_custom feature as optional

kairoaraujo:
It fixes the `add_task_id_to_custom` as an optional feature.

- Change the `add_task_id_to_custom` as optional
- Change the logic case the option is missing to force as None
- Add a check if the option is True
- Increate the UT for the API feature

### DIFF
--- a/repository_service_tuf_api/targets.py
+++ b/repository_service_tuf_api/targets.py
@@ -66,7 +66,7 @@ class AddPayload(BaseModel):
     """
 
     targets: List[Targets]
-    add_task_id_to_custom: bool
+    add_task_id_to_custom: Optional[bool]
 
     class Config:
         with open("tests/data_examples/targets/payload.json") as f:
@@ -108,17 +108,17 @@ def post(payload: AddPayload) -> Response:
         )
 
     task_id = get_task_id()
-    if payload.add_task_id_to_custom:
+    if payload.add_task_id_to_custom is True:
         new_targets: List[Targets] = []
         for target in payload.targets:
-            if target.info.custom is None:
-                target.info.custom = {}
+            if target.info.custom:
+                target.info.custom = {
+                    "added_by_task_id": task_id,
+                    **target.info.custom,
+                }
+            else:
+                target.info.custom = {"added_by_task_id": task_id}
 
-            # Add task_id info in custom while keeping the old custom
-            target.info.custom = {
-                "added_by_task_id": task_id,
-                **target.info.custom,
-            }
             new_targets.append(target)
 
         payload.targets = new_targets
@@ -126,7 +126,7 @@ def post(payload: AddPayload) -> Response:
     repository_metadata.apply_async(
         kwargs={
             "action": "add_targets",
-            "payload": payload.dict(by_alias=True),
+            "payload": payload.dict(by_alias=True, exclude_none=True),
         },
         task_id=task_id,
         queue="metadata_repository",
--- a/tests/unit/api/test_targets.py
+++ b/tests/unit/api/test_targets.py
@@ -70,7 +70,7 @@ class TestPostTargets:
             f_data = f.read()
 
         payload = json.loads(f_data)
-        payload["add_task_id_to_custom"] = False
+
         mocked_repository_metadata = pretend.stub(
             apply_async=pretend.call_recorder(lambda *a, **kw: None)
         )
@@ -110,7 +110,6 @@ class TestPostTargets:
         }
 
         # Add task_id info into custom as it will be done in the post function
-        expected_payload = {"targets": [], "add_task_id_to_custom": True}
         for target in payload["targets"]:
             if target["info"].get("custom") is None:
                 target["info"]["custom"] = {}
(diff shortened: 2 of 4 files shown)


kairoaraujo:
This PR fixes the current failure for the FT
https://github.com/vmware/repository-service-tuf/actions/runs/4013310896/jobs/6898595197

repository_service_tuf_api/targets.py
```diff
@@ -108,25 +108,25 @@ def post(payload: AddPayload) -> Response:
...
-            # Add task_id info in custom while keeping the old custom
-            target.info.custom = {
-                "added_by_task_id": task_id,
-                **target.info.custom,
-            }
+            if target.info.custom:
+                target.info.custom = {
+                    "added_by_task_id": task_id,
+                    **target.info.custom,
+                }
```
MVrachev, line 118: I just realized that nowhere it's specified that `custom` should be a dictionary.
The TUF specification defines `custom` [like this](https://theupdateframework.github.io/specification/latest/#custom):
```
An object. If defined, the elements and values of the CUSTOM object will be made
available to the client application. The format of the CUSTOM object is opaque to the
framework, which only needs to know that the "custom" attribute maps to an object.
The CUSTOM object may include version numbers, dependencies, requirements, or
any other data that the application wants to include to describe the file at TARGETPATH.
The application may use this information to guide download decisions.
```
In `python-tuf` when we access the `custom` field we don't say that the return type is a `Dict` or a `Map`:
https://github.com/theupdateframework/python-tuf/blob/a6460c60ba9ed3a609e43ab981ec5a941606d6f4/tuf/api/metadata.py#L1776

Maybe we should brainstorm what should we do if there is an existing `custom` which is not a `Dict` or a `Map`?

repository_service_tuf_api/targets.py
```diff
@@ -108,25 +108,25 @@ def post(payload: AddPayload) -> Response:
...
+
             new_targets.append(target)
 
         payload.targets = new_targets
 
     repository_metadata.apply_async(
         kwargs={
             "action": "add_targets",
-            "payload": payload.dict(by_alias=True),
+            "payload": payload.dict(by_alias=True, exclude_none=True),
```
MVrachev, line 129: Cool, never have I used this method in this way to exclude `None` values.

tests/unit/api/test_targets.py
```diff
@@ -50,6 +50,65 @@ def test_post(self, monkeypatch, test_client, token_headers):
...
+                    "action": "add_targets",
+                    "payload": payload,
+                },
+                task_id=fake_task_id,
+                queue="metadata_repository",
+                acks_late=True,
+            )
+        ]
+
+    def test_post_with_add_task_id_to_custom(
```
MVrachev, line 65: Why do we need this test when by you already test the case when `add_task_id_to_custom` is true in `test_post`?
Either rename `test_post` or give a comment that in this case we are using `add_task_id_to_custom` or make pass `add_task_id_to_custom` directly in the input `payload`.
Then you can add a simple test case only when `add_task_id_to_custom` is `False`. 

I am not a fan of a single test checking multiple scenarios as we did here.
We can have multiple tests for each case of `add_task_id_to_custom` as we usually do and if there is a repeating code we can always create helper setup functions.

tests/unit/api/test_targets.py
```diff
@@ -62,28 +121,27 @@ def test_post(self, monkeypatch, test_client, token_headers):
                 **target["info"]["custom"],
             }
             expected_payload["targets"].append(target)
-
         assert mocked_repository_metadata.apply_async.calls == [
             pretend.call(
                 kwargs={
                     "action": "add_targets",
-                    "payload": expected_payload,
+                    "payload": payload,
```
MVrachev, line 128: You can remove the `exected_payload` altogether in this test.

tests/unit/api/test_targets.py
```diff
@@ -62,28 +121,27 @@ def test_post(self, monkeypatch, test_client, token_headers):
...
                 },
                 task_id=fake_task_id,
                 queue="metadata_repository",
                 acks_late=True,
             )
         ]
 
-    def test_post_without_add_task_id_to_custom(
+    def test_post_with_add_task_id_to_custom_false(
```
MVrachev, line 136: This test is repeating what you do in `test_post_with_add_task_id_to_custom`.
Either you can remove the false case from `test_post_with_add_task_id_to_custom` or remove this test.
Another possible solution will be to keep this test with its original title but remove `add_task_id_to_custom` from the payload.
I am not sure if we have such a test without any value for `add_task_id_to_custom`.

MVrachev (changes requested):
Great job finding this bug!

I have some comments on the unit tests, when we agree on them this pr is ready to go. :))

repository_service_tuf_api/targets.py
MVrachev: We can open an issue if you agree there is a sense to discuss this.

kairoaraujo: The RSTUF defines some constraints in the Metadata, such as the delegations to succinct hash bin roles. The main reason is to make TUF easily adoptable and give some flexibility for use cases.
In that sense, using custom as a `Dict` helps the current TUF users that want to migrate to RSTUF once they need to define a key value for their specific custom value type.

tests/unit/api/test_targets.py
kairoaraujo: > I am not a fan of a single test checking multiple scenarios as we did here. We can have multiple tests for each case of `add_task_id_to_custom` as we usually do and if there is a repeating code we can always create helper setup functions.

In the RSTUF, we are adopting two standards for UT across all components
- One test function per scenario
- Be explicit about the test steps, avoiding nested functions to build scenarios or conditional helpers that will require tests and can bring false positives.

MVrachev (approved):

PR #241, opened 2023-01-28
Release v0.0.1a4

kairoaraujo:
- Updated the Copyright 2022-2023


MVrachev (approved):
We need to update the copyright in all of our files in the rest of the components and the umbrella.
Isn't there a way to write the copyright in such a way that this is not required each year?

Merging this pr and we can continue this discussion here, an issue or outside GitHub.

PR #244, opened 2023-01-30
fix make requirements and update dependencies

### LINKED ISSUE #238: Bug: make requirements does not update Pipfile.lock when Pipfile has been manually
### What happened?

Running make requirements does not update the requirement files (requirements.txt and requirements-dev.txt) correctly if we have manually edited the Pipfile. This could happen if we are manually pinning a dependency or specifying some optional dependency.
Locking first could protect us from having outdated requirement files if we manually edit the Pipfile and would be a nice-to-have extra precaution.

### What steps did you take?

1. Update the Pipfile by adding a new package
2. Run make requirements to update the requirement files

We expect the requirements files to include the latest changes made to the Pipfile.
This does not happen since we need to lock before running:

```
pipenv requirements > requirements.txt
pipenv requirements --dev > requirements-dev.txt
```
so we could instead have:

```
pipenv lock
pipenv requirements > requirements.txt
pipenv requirements --dev > requirements-dev.txt
```
This is because the Pipfile.lock is used to generate the requirement files

### Relevant log output

_No response_

### Code of Conduct

kairoaraujo:
- fix `make requirements`
- update the API requirements and Pipfile

Closes #238

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,7 +24,7 @@ jobs:
       run: pip install tox tox-gh-actions
 
     - name: Run Python tests
-      run: tox
+      run: tox -r
 
     - name: Codecov
       uses: codecov/codecov-action@d9f34f8cd5cb3b3eb79b3e4b5dae3a16df499a70
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,12 +15,12 @@ repos:
   - id: flake8
     exclude: repository_service_tuf_api/__init__.py|venv|.venv|setting.py|.git|.tox|dist|docs|/*lib/python*|/*egg|build|tools
 - repo: https://github.com/PyCQA/isort
-  rev: '5.10.1'
+  rev: '5.12.0'
   hooks:
   - id: isort
     args: [-l79, --profile, black, --check, --diff]
 - repo: https://github.com/psf/black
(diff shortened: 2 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -24,7 +24,7 @@ jobs:
       run: pip install tox tox-gh-actions
 
     - name: Run Python tests
-      run: tox
+      run: tox -r
```
KAUTH, line 27: Nit:
```suggestion
      make tests
```

KAUTH (approved):
LGTM, just a small nit

.github/workflows/ci.yml
MVrachev: Does the GitHub workflow see `MAKEFILE`?

MVrachev (approved):

.github/workflows/ci.yml
kairoaraujo: Yes, it does see it. 
But we need to refine and decide better about the Makefile structure. In our dev docs, we share the tox in our CI using make in some parts. It's the second time we think about it ([first one](https://github.com/vmware/repository-service-tuf-cli/pull/110#pullrequestreview-1159892929)). What is good.😄 
I will merge and we can discuss it more in the below issue:
- https://github.com/vmware/repository-service-tuf/issues/229

Issue #17, opened 2022-07-06
Refactoring names in the implementation for API Models and implementation/endpoint functions


kairoaraujo:
Improve the names for the API Models and the functions names.

``Response`` ->  ``BootstrapResponse``
``get``  -> ``get_bootstrap``

KAUTH:
Doing some cleanup:
- `BootstrapResponse` was changed to `BootstrapGetResponse` and `BootstrapPostResponse` accordingly in https://github.com/repository-service-tuf/repository-service-tuf-api/commit/d24b69bfd8d54c931d17bf6e40a4968ccbea4807#diff-8ce4883d268e87f0aaed768dbec3413cbbbdb442ca8788479f242e88665f0484
- `get_bootstrap` is also named as desired

@kairoaraujo I believe we can close this issue.

PR #250, opened 2023-02-01
add API endpoint `/api/v1/targets/publish`

### LINKED ISSUE #242: Feature: Add endpoint to trigger `publish_targets` in the Rest API
### What do you want to share with us?

Add to Rest API an endpoint to trigger `publish_targets`.

Implementation suggestion
- Add `/api/v1/targets/publish` with the method POST
- It will send to broker a task with the action `publish_targets` and return `ACCEPTED` and the task id (follows the same task request for add targets 

### References

- https://github.com/vmware/repository-service-tuf-worker/issues/175

### Code of Conduct

kairoaraujo:
It will send to broker a task with the action `publish_targets` and return HTTP 202 and the task id (follows the same task request for add targets).

Closes #242

### DIFF
--- a/repository_service_tuf_api/api/targets.py
+++ b/repository_service_tuf_api/api/targets.py
@@ -48,3 +48,23 @@ def delete(
     _user=Security(validate_token, scopes=[SCOPES_NAMES.delete_targets.value]),
 ):
     return targets.delete(payload)
+
+
+@router.post(
+    "/publish/",
+    summary=(
+        "Submit a task to publish targets."
+        f"Scope: {SCOPES_NAMES.write_targets.value}"
+    ),
+    description=(
+        "Trigger a task to publish targets not yet published from the "
+        "RSTUF Database"
+    ),
+    response_model=targets.Response,
+    response_model_exclude_none=True,
+    status_code=status.HTTP_202_ACCEPTED,
+)
+def post_publish_targets(
+    _user=Security(validate_token, scopes=[SCOPES_NAMES.write_targets.value]),
+):
+    return targets.post_publish_targets()
--- a/repository_service_tuf_api/targets.py
+++ b/repository_service_tuf_api/targets.py
@@ -17,7 +17,7 @@ from repository_service_tuf_api.metadata import (
 
 
 class ResponseData(BaseModel):
-    targets: List[str]
+    targets: Optional[List[str]]
     task_id: str
     last_update: datetime
 
@@ -171,4 +171,27 @@ def delete(payload: DeletePayload) -> Response:
     }
     return Response(
         data=data, message="Remove Target(s) successfully submitted."
+    )
+
+
+def post_publish_targets() -> Response:
+    task_id = get_task_id()
+    repository_metadata.apply_async(
+        kwargs={
+            "action": "publish_targets",
+            "payload": None,
+        },
+        task_id=task_id,
+        queue="rstuf_internals",
+        acks_late=True,
+    )
+
+    data = {
+        "targets": None,
+        "task_id": task_id,
+        "last_update": datetime.now(),
+    }
+
+    return Response(
+        data=data, message="Publish targets successfully submitted."
     )
--- a/repository_service_tuf_api/targets.py
+++ b/repository_service_tuf_api/targets.py
@@ -17,7 +17,7 @@ from repository_service_tuf_api.metadata import (
 
 
 class ResponseData(BaseModel):
-    targets: Optional[List[str]]
+    targets: List[str]
     task_id: str
     last_update: datetime
 
@@ -187,7 +187,7 @@ def post_publish_targets() -> Response:
     )
 
     data = {
-        "targets": None,
+        "targets": [],
         "task_id": task_id,
         "last_update": datetime.now(),
     }
--- a/tests/unit/api/test_targets.py
+++ b/tests/unit/api/test_targets.py
@@ -353,6 +353,7 @@ class TestPostTargetsPublish:
         assert response.status_code == status.HTTP_202_ACCEPTED
         assert response.json() == {
             "data": {
+                "targets": [],
                 "task_id": fake_task_id,
                 "last_update": "2019-06-16T09:05:01",
             },
--- a/tests/unit/api/test_targets.py
+++ b/tests/unit/api/test_targets.py
@@ -18,7 +18,7 @@ class TestPostTargets:
         payload = json.loads(f_data)
 
         mocked_repository_metadata = pretend.stub(
-            apply_async=pretend.call_recorder(lambda *a, **kw: None)
+            apply_async=pretend.call_recorder(lambda **kw: None)
         )
         monkeypatch.setattr(
             "repository_service_tuf_api.targets.is_bootstrap_done",
@@ -72,7 +72,7 @@ class TestPostTargets:
         payload = json.loads(f_data)
 
         mocked_repository_metadata = pretend.stub(
-            apply_async=pretend.call_recorder(lambda *a, **kw: None)
+            apply_async=pretend.call_recorder(lambda **kw: None)
         )
         monkeypatch.setattr(
             "repository_service_tuf_api.targets.is_bootstrap_done",
@@ -219,7 +219,7 @@ class TestDeleteTargets:
             lambda: True,
         )
         mocked_repository_metadata = pretend.stub(
-            apply_async=pretend.call_recorder(lambda *a, **kw: None)
+            apply_async=pretend.call_recorder(lambda **kw: None)
         )
         monkeypatch.setattr(
             "repository_service_tuf_api.targets.repository_metadata",
@@ -331,7 +331,7 @@ class TestPostTargetsPublish:
         url = "/api/v1/targets/publish/"
 
         mocked_repository_metadata = pretend.stub(
-            apply_async=pretend.call_recorder(lambda *a, **kw: None)
+            apply_async=pretend.call_recorder(lambda **kw: None)
         )
         monkeypatch.setattr(
             "repository_service_tuf_api.targets.repository_metadata",
--- a/tests/unit/api/test_targets.py
+++ b/tests/unit/api/test_targets.py
@@ -324,3 +324,48 @@ class TestDeleteTargets:
         assert response.json() == {
             "detail": {"error": "scope 'delete:targets' not allowed"}
         }
+
+
+class TestPostTargetsPublish:
+    def test_post_publish(self, monkeypatch, test_client, token_headers):
+        url = "/api/v1/targets/publish/"
+
+        mocked_repository_metadata = pretend.stub(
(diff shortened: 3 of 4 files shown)


repository_service_tuf_api/api/targets.py
```diff
@@ -48,3 +48,23 @@ def delete(
     _user=Security(validate_token, scopes=[SCOPES_NAMES.delete_targets.value]),
 ):
     return targets.delete(payload)
+
+
+@router.post(
+    "/publish/",
```
MVrachev, line 54: I see that not all of the methods have this attribute set to anything different than `/`.
Am I right in my observation that the endpoint names are constructed like that if they are an f-string:
`f"api/{routert_tag}/{router_prefix}/{method_suffix}"` ?

repository_service_tuf_api/api/targets.py
```diff
@@ -48,3 +48,23 @@ def delete(
...
+    summary=(
+        "Submit a task to publish targets."
+        f"Scope: {SCOPES_NAMES.write_targets.value}"
+    ),
+    description=(
+        "Trigger a task to publish targets not yet published from the "
+        "RSTUF Database"
```
MVrachev, line 61: When I read this description I asked myself: how will the user know who is the current batch of non-published targets?

repository_service_tuf_api/api/targets.py
```diff
@@ -48,3 +48,23 @@ def delete(
...
+    ),
+    response_model=targets.Response,
+    response_model_exclude_none=True,
+    status_code=status.HTTP_202_ACCEPTED,
```
MVrachev, line 65: When is this status code given?
The default response code when an allowed API call is made?

repository_service_tuf_api/api/targets.py
```diff
@@ -48,3 +48,23 @@ def delete(
...
+)
+def post_publish_targets(
```
MVrachev, line 67: Why not call this method `publish_targets` instead?
Same for `targets.post_publish_targets()` below.

repository_service_tuf_api/targets.py
```diff
@@ -17,7 +17,7 @@
 
 
 class ResponseData(BaseModel):
-    targets: List[str]
+    targets: Optional[List[str]]
```
MVrachev, line 20: I see why you make this list optional, but it shouldn't be optional for `add/remove targets`.
Should we use other structures for those methods?
Another option will be to not do anything if this list is `None`, but if that's the case we should make sure we test it.

tests/unit/api/test_targets.py
```diff
@@ -324,3 +324,48 @@ def test_post_forbidden_user_incorrect_scope_token(self, test_client):
...
             "detail": {"error": "scope 'delete:targets' not allowed"}
         }
+
+
+class TestPostTargetsPublish:
+    def test_post_publish(self, monkeypatch, test_client, token_headers):
+        url = "/api/v1/targets/publish/"
+
+        mocked_repository_metadata = pretend.stub(
+            apply_async=pretend.call_recorder(lambda *a, **kw: None)
```
MVrachev, line 334: Wondering when you mock when do you need both `*a` and `**kw` in the lambda function?
Right now I see that for `apply_async` you use keyword arguments only. 
Why do you also need `*a`?

repository_service_tuf_api/targets.py
```diff
@@ -172,3 +172,26 @@ def delete(payload: DeletePayload) -> Response:
...
+
+def post_publish_targets() -> Response:
+    task_id = get_task_id()
+    repository_metadata.apply_async(
+        kwargs={
+            "action": "publish_targets",
+            "payload": None,
+        },
+        task_id=task_id,
+        queue="rstuf_internals",
```
MVrachev, line 185: That's the name of the message queue name where the task should go, right?

repository_service_tuf_api/targets.py
```diff
@@ -172,3 +172,26 @@ def delete(payload: DeletePayload) -> Response:
...
+        acks_late=True,
```
MVrachev, line 186: What does `acks_late` does?

MVrachev (commented):
As this is my first review on a pr adding a new endpoint I have some questions.

repository_service_tuf_api/api/targets.py
kairoaraujo: The `app.py` defines the `rstuf_app` "/"
https://github.com/vmware/repository-service-tuf-api/blob/3392117cdca4488158b02958e1601796f8766b0a/app.py#L17-L20

The `api_v1` also defines the "/api/v1/"

https://github.com/vmware/repository-service-tuf-api/blob/3392117cdca4488158b02958e1601796f8766b0a/app.py#L23

- What allows you to have multiple API versions (v2, v3...) to keep the backward compatibility with old versions of API Clients

In the `repository_service_tuf_api/api/` we implement the routes, enabling us to have multiple implementations if needed and import them with specific versions or multiple

https://github.com/vmware/repository-service-tuf-api/blob/3392117cdca4488158b02958e1601796f8766b0a/app.py#L10-L15

repository_service_tuf_api/api/targets.py
kairoaraujo: Yes, there is no way to know what is not published. The user can control it by themself from:
- the requests and the [status of the tasks](https://github.com/vmware/repository-service-tuf-worker/issues/190) ([if the targets are published in the same request](https://github.com/vmware/repository-service-tuf-worker/issues/175#issuecomment-1410644318))
- if they don't publish in the same request (above), by using this endpoint, we are implementing

I think it is more about documenting in the Guide: 
- https://github.com/vmware/repository-service-tuf/issues/227

repository_service_tuf_api/api/targets.py
kairoaraujo: When the the request is returned, this is default HTTP status code.
For example gets an exception, automatically will be HTTP 500 or the request is not correct it will raise 400.
Specific return codes can also be done, but not in this case.
It is the https://fastapi.tiangolo.com framework.

repository_service_tuf_api/api/targets.py
kairoaraujo: I won't be changing it now because I'm following to have the method name in the functions and its is nothins exposed to the users.

But, there is an old issue about refactoring this functions name. (I will start renaming the issue from July 2022)
- #17

repository_service_tuf_api/targets.py
kairoaraujo: You're right, actually I don't need do it. FastAPI does it.
I can just send a empty list in `publish_targets`.

repository_service_tuf_api/targets.py
kairoaraujo: Yes, we have [two queues in RSTUF.](https://repository-service-tuf.readthedocs.io/projects/rstuf-worker/en/latest/devel/index.html#adding-removing-targets)
The publish targets are used in the `rstuf_internals`

repository_service_tuf_api/api/targets.py
MVrachev: > * What allows you to have multiple API versions (v2, v3...) to keep the backward compatibility with old versions of API Clients
> 
> In the `repository_service_tuf_api/api/` we implement the routes, enabling us to have multiple implementations if needed and import them with specific versions or multiple
> 
> https://github.com/vmware/repository-service-tuf-api/blob/3392117cdca4488158b02958e1601796f8766b0a/app.py#L10-L15

Cool, thanks for mentioning that part as well.

MVrachev: So, the naming is 
`f"{api_v1.prefix}/{router_prefix}{method_suffix}"`?

Then, what is the role of the `tags` and `response` attributes for the APIRouter?
It seems they are not used in the naming.

repository_service_tuf_api/api/targets.py
MVrachev: That's okay if I just added a batch of targets without publishing them, but if someone is serious about this feature and wants big batches of targets before publishing them it's not scaling well.
Do you think we need an endpoint or other way to show this?
Maybe we can mention other non-published changes - added or removed targets in the response to a user's requests to add or remove targets without publishing them?

repository_service_tuf_api/api/targets.py
MVrachev: Yea, but why add new methods to be refactored when this is a new feature not been introduced? 
Do you follow a certain format now?

repository_service_tuf_api/targets.py
MVrachev: Cool that we have documentation about this.

repository_service_tuf_api/api/targets.py
kairoaraujo: > So, the naming is `f"{api_v1.prefix}/{router_prefix}{method_suffix}"`?

`f"{api_v1.prefix}/{router.prefix}/{router.<method>.path}"`

> Then, what is the role of the `tags` and `response` attributes for the APIRouter? It seems they are not used in the naming.

Tags are used in the Documentation. The same router can be part of different tags, so you can reuse the same router in multiple API prefixes. It is more on the[ FastAPI framework details.](https://fastapi.tiangolo.com/advanced/)

Simulation.
Screenshot 2023-02-03 at 12 46 42

The responses are also related to swagger documentation.
https://fastapi.tiangolo.com/tutorial/response-status-code/
https://fastapi.tiangolo.com/advanced/additional-responses/

repository_service_tuf_api/api/targets.py
kairoaraujo: > That's okay if I just added a batch of targets without publishing them, but if someone is serious about this feature and wants big batches of targets before publishing them it's not scaling well.
> Do you think we need an endpoint or other way to show this?

In that case, We don't need an endpoint for it. We should give an additional optional parameter in the payload for this method/endpoint, passing which delegated roles they want to publish. I don't think we should go deep on it now. It can be done in another interaction upon request as a new feature or feature enhancement.

> Maybe we can mention other non-published changes - added or removed targets in the response to a user's requests to add or remove targets without publishing them?

We do it for the add/removing targets. We return the published targets.
Doing this control will impact the performance computation, as we must interact in the middle. Let's start simple, and we can improve if there is a user case/request for it.

repository_service_tuf_api/api/targets.py
kairoaraujo: Because we need to decide on the format/standard and apply it consistently, we have this issue to continue the discussion there.
It will not have a significant improvement as this is internal functionality. It doesn't solve or introduce a technical-debit problem refactoring this name and also doesn't unblock our roadmap.
A small PR can solve it with a defined standard for naming the FastAPI functions for endpoints.

MVrachev: Okay, let's follow in a discussion about similar changes there in this issue.

Issue #248, opened 2023-02-01
feature: add optional field `publish_targets` in add/remove payload


kairoaraujo:
### What do you want to share with us?

- This feature is was originated in https://github.com/vmware/repository-service-tuf-worker/issues/175 discussion
- Add an optional field in the add and remove payloads
  - name: `publish_targets`
  - default: True
 

### References

- https://github.com/vmware/repository-service-tuf-worker/issues/198
- https://github.com/vmware/repository-service-tuf-worker/issues/175

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #254, opened 2023-02-08
Add publish targets option

MVrachev:
Add an optional field called "publish_targets" in the add/remove
targets API calls. If no value is provided or it's set to true,
publishing of targets will happen the same as now: a subtask will
be started that on completion will have published the modified targets.
If it's set to false, publishing of the targets file changes won't be
published (meaning there won't be any changes in the metadata files
as well) until specifically requested with the newly introduced
API call "api/v1/targets/publish".

This feature is was originated in Task:
https://github.com/vmware/repository-service-tuf-worker/issues/175
discussion.

Additionally, give add_task_id_to_custom a default value.

Close https://github.com/vmware/repository-service-tuf-api/issues/248

### DIFF
--- a/repository_service_tuf_api/targets.py
+++ b/repository_service_tuf_api/targets.py
@@ -66,7 +66,10 @@ class AddPayload(BaseModel):
     """
 
     targets: List[Targets]
-    add_task_id_to_custom: Optional[bool]
+    add_task_id_to_custom: Optional[bool] = Field(
+        default=False,
+        description="Whether to add the id of the task in custom",
+    )
     publish_targets: Optional[bool] = Field(
         default=True, description="Whether to publish the targets"
     )
--- a/repository_service_tuf_api/targets.py
+++ b/repository_service_tuf_api/targets.py
@@ -7,7 +7,7 @@ from datetime import datetime
 from typing import Any, Dict, List, Optional
 
 from fastapi import HTTPException, status
-from pydantic import BaseModel
+from pydantic import BaseModel, Field
 
 from repository_service_tuf_api.metadata import (
     get_task_id,
@@ -67,6 +67,9 @@ class AddPayload(BaseModel):
 
     targets: List[Targets]
     add_task_id_to_custom: Optional[bool]
+    publish_targets: Optional[bool] = Field(
+        default=True, description="Whether to publish the targets"
+    )
 
     class Config:
         with open("tests/data_examples/targets/payload.json") as f:
@@ -81,6 +84,9 @@ class DeletePayload(BaseModel):
     """
 
     targets: List[str]
+    publish_targets: Optional[bool] = Field(
+        default=True, description="Whether to publish the targets changes"
+    )
 
     class Config:
         payload = {
@@ -132,12 +138,17 @@ def post(payload: AddPayload) -> Response:
         queue="metadata_repository",
         acks_late=True,
     )
+
+    message = "Target(s) successfully submitted."
+    if payload.publish_targets is False:
+        message += " Publishing will be skipped."
+
     data = {
         "targets": [target.path for target in payload.targets],
         "task_id": task_id,
         "last_update": datetime.now(),
     }
-    return Response(data=data, message="Target(s) successfully submitted.")
+    return Response(data=data, message=message)
 
 
 def delete(payload: DeletePayload) -> Response:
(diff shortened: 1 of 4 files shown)


tests/data_examples/targets/payload.json
```diff
@@ -19,5 +19,7 @@
             },
             "path": "file2.tar.gz"
         }
-    ]
+    ],
+    "publish_targets": true,
+    "add_task_id_to_custom": false
```
kairoaraujo, line 24: I think we should remove the optional fields from the example `payload.json`.
It will force us to test the default value scenario. If someone removes the field as Optional or changes the default we can notice it in the tests.

kairoaraujo (changes requested):
I suggest removing the `docs/source/_static/` from this PR.
I also have a small comment about the tests, but it looks good!

tests/data_examples/targets/payload.json
MVrachev: I added it in the example because otherwise, I had to change multiple tests and add a value to those fields because using `pydantic.Field()` for them with a default value means they will always have a value no matter if we give them inside the `payload.json`.
So if we don't want them in this example I would have to add them manually in the payload in the tests.

repository_service_tuf_api/targets.py
```diff
@@ -66,7 +66,13 @@ class AddPayload(BaseModel):
     """
 
     targets: List[Targets]
-    add_task_id_to_custom: Optional[bool]
+    add_task_id_to_custom: Optional[bool] = Field(
+        default=False,
+        description="Whether to add the id of the task in custom",
+    )
+    publish_targets: Optional[bool] = Field(
```
KAUTH, line 73: Hey, just passing by here, a small comment. Shouldn't the type be `bool` (instead of `Optional[bool]`) since we give a default. Same for [line 90](https://github.com/vmware/repository-service-tuf-api/pull/254/files#diff-04e45fdfaa48b6ef3b8855ee0940eaeccf5b9010e5613a74fdaf617b351da35aR90)

MVrachev: Making it optional means that the user is not required to give any value to the `publish_targets` attribute.
If no value is given, then pydantic will make sure to populate it.
We do that to escape the case when we have `publish_targets` None and we have missed handling it.

KAUTH: You can also make it optional by setting it as `bool = Field(default=True, description="Whether to publish the targets")`, I tested both ways. What I like about the aforementioned typing is that when you use it later on, for example in:
```python
def post(payload: AddPayload) -> Response:
...
...
    if payload.publish_targets is False:
...
``` 
you can easily tell that `payload.publish_targets` is a `bool` (even when not provided by the user, since it has a default) and not `None`.

In any case, I tested in `mypy` with both typing and it doesn't complain, so I guess it's fine to leave it as is.

KAUTH (changes requested):
Could you also update the example in `tests/data_examples/targets/payload.json` to include `add_task_id_to_custom` and `publish_targets`?

kairoaraujo:
> Could you also update the example in `tests/data_examples/targets/payload.json` to include `add_task_id_to_custom` and `publish_targets`?
 
As I mentioned [above](https://github.com/vmware/repository-service-tuf-api/pull/254#discussion_r1101404361), I'm not in favor of adding the `add_task_id_to_custom` and `publish_targets` in the `payload.json` for the tests. IMHO, we should keep only the required fields. The reasons are:

- We are also testing if nobody changed the parameter to not optional
- We are testing the real user flow, inserting the parameters in the payload.

KAUTH:
> > Could you also update the example in `tests/data_examples/targets/payload.json` to include `add_task_id_to_custom` and `publish_targets`?
> 
> As I mentioned [above](https://github.com/vmware/repository-service-tuf-api/pull/254#discussion_r1101404361), I'm not in favor of adding the `add_task_id_to_custom` and `publish_targets` in the `payload.json` for the tests. IMHO, we should keep only the required fields. The reasons are:
> 
>     * We are also testing if nobody changed the parameter to not optional
> 
>     * We are testing the real user flow, inserting the parameters in the payload.

Oh sorry, although I read the previous comment I forgot about it.
You are right, we could have had the optional fields and then pop them to test if the parameters are optional, but then it would not be as close to the real user flow.

And I see, this way you are coupling documentation and testing making sure that whatever is documented is the same thing being tested, which is pretty nice!

repository_service_tuf_api/targets.py
MVrachev: You have a point here. Given that we know that `pydantic` will populate the value it's not correct to say it's an optional bool, it's a bool.
Will update it to make it clear.

kairoaraujo (approved):
LGTM

PR #258, opened 2023-02-12
Add dockerignore file

kapsalis:
Add dockerignore file to exclude .tox and venv directories when sending build context to the Docker deamon. These dirs may have a big size and they are not needed during the docker image build process.

Signed-off-by: Michalis Kapsalakis <kapsalis1989@gmail.com>


MVrachev:
Thanks for your contribution @kapsalis!
Can you also add `.env` and `.vscode` directories to the ignore list?
Those are directories I regularly use locally and will appreciate that.

kairoaraujo:
Thank you, @kapsalis. 

I'm in favor of having .`dockerignore`, but we don't add extra files to the container image.
We carefully add the required files if you see the `Dockerfile`. We don't do `ADD . <path>`

https://github.com/vmware/repository-service-tuf-api/blob/09191a8f9c4bf1b5167708ba56dde32e381dd9da/Dockerfile#L29-L35

We could also remove the tests from the Dockerfile.

kapsalis:
@kairoaraujo that's true, but dockerignore helps to reduce the time needed to start building the image. It is not related to the files will be final in the image (of course if we exclude something that is needed, the build will fail), but on what is copied on the docker daemon.

On my machine, the two folders, which are ignored now, were around 900MB. It took around 10-15 seconds to start the build. After ignoring, the context was 20MB, and build started instantly.

I would say that it is mainly for local development speed up.

kairoaraujo:
> I would say that it is mainly for local development speed up.

That's true too. I'm in favor of using the `.dockerignore`.

btw, are using the `make run-dev` for running your local development environment? It makes an automatic reload for changes in the current container you're developing. Does it help the `make run-dev` somehow? 🤔

kapsalis:
Hi again! Yes, I use the `make run-dev`. It is really helpful to build fast and run the container.

Dockerignore will help only during the build. So, when you run the script. Then, it is not related to any changes to the code that reload the app, which still happens

MVrachev (approved):
@kairoaraujo I am a little unsure do you agree with those changes?
Sounds like they are what you suggest as well given that @kapsalis is using `.dockerignore` as you suggested.

I am happy with the changes and thanks @kapsalis for providing more info about the speed up regarding the docker image startup.

@kapsalis if we merge this pr here could you do the same for [RSTUF Worker](https://github.com/vmware/repository-service-tuf-worker)? 
I ask you because as you can start an RSTUF instance from there as well

kapsalis:
Sure! I am happy to do it on the workers too.

Just to clarify, the change speeds up the image build startup, not the container startup.

Issue #65, opened 2022-08-19
Revoke token


kairoaraujo:
Implement a way to revoke a token (?)

MVrachev:
No longer viable as we are removing authentication and authorization of RSTUF: https://github.com/repository-service-tuf/repository-service-tuf/issues/495

PR #264, opened 2023-02-19
Kauth/remove unused keys from access token

### LINKED ISSUE #263: Task: Remove `sub`, `username` and `password` from returned `access_token`
### What is the task about?

The values of `sub`, `username` and `password` that are encoded in the `access_token` that is returned to the user for authentication/authorization (POST `/api/v1/token` or `/api/v1/token/new`) are not used to validate the token ([here](https://github.com/vmware/repository-service-tuf-api/blob/main/repository_service_tuf_api/token.py#L129)) or for any other reason, so they should be removed.
*The `username` could be potentially used in the future for authorization, but since now we only have one admin user it is not needed.*

Example payload of `access_token` now:
Token: eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJ1c2VyXzFfNjZjNzEyMGQxM2Q4NDRlY2E1ZDczNDA4NjBkNWM1OTIiLCJ1c2VybmFtZSI6ImFkbWluIiwicGFzc3dvcmQiOiJiJyQyYiQxMiRFZnltS1Y5RWdEeVZoLy5RSnEwaEl1NHdiV2J4M0JXa3VhbEFvLmRaUkpSNW5PRENvRjhCaSciLCJzY29wZXMiOlsicmVhZDpib290c3RyYXAiLCJyZWFkOnNldHRpbmdzIiwicmVhZDp0YXNrcyIsInJlYWQ6dG9rZW4iLCJ3cml0ZTpib290c3RyYXAiLCJ3cml0ZTp0YXJnZXRzIiwid3JpdGU6dG9rZW4iLCJkZWxldGU6dGFyZ2V0cyJdLCJleHAiOjIwMzY4Mzk1NjJ9.ko614QA9Ud3r3Y64O_a6jTIzxlm6PmI1M-hhuZNHZ7E
Decoded token payload:
```json
{
  "sub": "user_1_66c7120d13d844eca5d7340860d5c592",
  "username": "admin",
  "password": "b'$2b$12$EfymKV9EgDyVh/.QJq0hIu4wbWbx3BWkualAo.dZRJR5nODCoF8Bi'",
  "scopes": [
    "read:bootstrap",
    "read:settings",
    "read:tasks",
    "read:token",
    "write:bootstrap",
    "write:targets",
    "write:token",
    "delete:targets"
  ],
  "exp": 2036839562
}
```

Proposed decoded token payload will look like:
```json
{
  "scopes": [
    "read:bootstrap",
    "read:settings",
    "read:tasks",
    "read:token",
    "write:bootstrap",
    "write:targets",
    "write:token",
    "delete:targets"
  ],
  "exp": 2036839562
}
```

Let me know what you think, I'm opening a PR 😊

### Parent feature

_No response_

### References

_No response_

### Code of Conduct

KAUTH:
The values of `sub`, `username` and `password` that are encoded in the
`access_token` that is returned to the user for authentication/authorization
(POST `/api/v1/token` or `/api/v1/token/new`) are not used to validate the token
or for any other reason, so they are removed.

Closes #263

Signed-off-by: Konstantinos Papadopoulos <konpap1996@yahoo.com>

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -263,45 +263,45 @@
         },
         "pydantic": {
             "hashes": [
-                "sha256:05a81b006be15655b2a1bae5faa4280cf7c81d0e09fcb49b342ebf826abe5a72",
-                "sha256:0b53e1d41e97063d51a02821b80538053ee4608b9a181c1005441f1673c55423",
-                "sha256:2b3ce5f16deb45c472dde1a0ee05619298c864a20cded09c4edd820e1454129f",
-                "sha256:2e82a6d37a95e0b1b42b82ab340ada3963aea1317fd7f888bb6b9dfbf4fff57c",
-                "sha256:301d626a59edbe5dfb48fcae245896379a450d04baeed50ef40d8199f2733b06",
-                "sha256:39f4a73e5342b25c2959529f07f026ef58147249f9b7431e1ba8414a36761f53",
-                "sha256:4948f264678c703f3877d1c8877c4e3b2e12e549c57795107f08cf70c6ec7774",
-                "sha256:4b05697738e7d2040696b0a66d9f0a10bec0efa1883ca75ee9e55baf511909d6",
-                "sha256:51bdeb10d2db0f288e71d49c9cefa609bca271720ecd0c58009bd7504a0c464c",
-                "sha256:55b1625899acd33229c4352ce0ae54038529b412bd51c4915349b49ca575258f",
-                "sha256:572066051eeac73d23f95ba9a71349c42a3e05999d0ee1572b7860235b850cc6",
-                "sha256:6a05a9db1ef5be0fe63e988f9617ca2551013f55000289c671f71ec16f4985e3",
-                "sha256:6dc1cc241440ed7ca9ab59d9929075445da6b7c94ced281b3dd4cfe6c8cff817",
-                "sha256:6e7124d6855b2780611d9f5e1e145e86667eaa3bd9459192c8dc1a097f5e9903",
-                "sha256:75d52162fe6b2b55964fbb0af2ee58e99791a3138588c482572bb6087953113a",
-                "sha256:78cec42b95dbb500a1f7120bdf95c401f6abb616bbe8785ef09887306792e66e",
-                "sha256:7feb6a2d401f4d6863050f58325b8d99c1e56f4512d98b11ac64ad1751dc647d",
-                "sha256:8775d4ef5e7299a2f4699501077a0defdaac5b6c4321173bcb0f3c496fbadf85",
-                "sha256:887ca463c3bc47103c123bc06919c86720e80e1214aab79e9b779cda0ff92a00",
-                "sha256:9193d4f4ee8feca58bc56c8306bcb820f5c7905fd919e0750acdeeeef0615b28",
-                "sha256:983e720704431a6573d626b00662eb78a07148c9115129f9b4351091ec95ecc3",
-                "sha256:990406d226dea0e8f25f643b370224771878142155b879784ce89f633541a024",
-                "sha256:9cbdc268a62d9a98c56e2452d6c41c0263d64a2009aac69246486f01b4f594c4",
-                "sha256:a48f1953c4a1d9bd0b5167ac50da9a79f6072c63c4cef4cf2a3736994903583e",
-                "sha256:a9a6747cac06c2beb466064dda999a13176b23535e4c496c9d48e6406f92d42d",
-                "sha256:a9f2de23bec87ff306aef658384b02aa7c32389766af3c5dee9ce33e80222dfa",
-                "sha256:b5635de53e6686fe7a44b5cf25fcc419a0d5e5c1a1efe73d49d48fe7586db854",
-                "sha256:b6f9d649892a6f54a39ed56b8dfd5e08b5f3be5f893da430bed76975f3735d15",
-                "sha256:b9a3859f24eb4e097502a3be1fb4b2abb79b6103dd9e2e0edb70613a4459a648",
-                "sha256:cd8702c5142afda03dc2b1ee6bc358b62b3735b2cce53fc77b31ca9f728e4bc8",
-                "sha256:d7b5a3821225f5c43496c324b0d6875fde910a1c2933d726a743ce328fbb2a8c",
-                "sha256:d88c4c0e5c5dfd05092a4b271282ef0588e5f4aaf345778056fc5259ba098857",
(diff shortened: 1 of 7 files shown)


kapsalis:
Hi @KAUTH . You still need the `username` key in the access token, in order to be able to authenticate the user. See the `api/v1/token/new`. It [validates](https://github.com/vmware/repository-service-tuf-api/blob/main/repository_service_tuf_api/api/token.py#L60) the token, and it returns its body. Then from this body, it uses the `username` field to search for the user (see [here](https://github.com/vmware/repository-service-tuf-api/blob/main/repository_service_tuf_api/token.py#L193)).

I totally agree that the password must not be part of the access token. Also, instead of using the `username`, the `sub` could be used with a format that can redirect to the user. Either with the username or by using uuids for the user id and use it in the sub.

kairoaraujo:
This is an awesome PR, @KAUTH. 🎉 
Thank you for the important suggestion @kapsalis 🚀

MVrachev:
> Hi @KAUTH . You still need the `username` key in the access token, in order to be able to authenticate the user. See the `api/v1/token/new`. It [validates](https://github.com/vmware/repository-service-tuf-api/blob/main/repository_service_tuf_api/api/token.py#L60) the token, and it returns its body. Then from this body, it uses the `username` field to search for the user (see [here](https://github.com/vmware/repository-service-tuf-api/blob/main/repository_service_tuf_api/token.py#L193)).

If @KAUTH didn't notice he still needs the `username` of a token doesn't this mean we don't test this?
We need to add a test to make sure that changes in the token body are noticed and we won't miss something like this.

Great catch @kapsalis!

kapsalis:
You are right @MVrachev . In [these API tests](https://github.com/vmware/repository-service-tuf-api/blob/main/tests/unit/api/test_token.py#L154), the `get_by_username` is mocked. As a result, @KAUTH could not see the issue.

KAUTH:
Thank you all for the very useful input guys! :100:
I'll update with changes

kapsalis:
@KAUTH indeed the `/targets` and `/tasks` do not use the `username` from the token. However, the `/token/new` endpoint uses the `username` to check if the user has the requested scopes. In my humble opinion, all the endpoints should check if the user exists and then check the scopes, so all the endpoints must use the `username`. We can unify this in the PR for the rstuf_auth service.

By the way, I think that even sharing the `username` in the token is not a good and secure approach. The token should not expose any sensitive information of the user (even on this small scale). My suggestion is to change the `user.id` type from `int` to `str` and generate a `uuid4().hex` value. In the token, we include the `id` as `sub`, and we use it when we authenticate the user with the token in the `rstuf_auth` service.

KAUTH:
Hey @kapsalis thanks for taking another look at the PR. I have a couple of questions:

>  However, the `/token/new` endpoint uses the `username` to check if the user has the requested scopes

I tried this with the current PR (without the `username`) and it works. Have you tried it with the current code in this PR and found that it doesn't?

>  My suggestion is to change the user.id type from int to str and generate a uuid4().hex value. In the token, we include the id as sub, and we use it when we authenticate the user with the token in the rstuf_auth service.

Maybe this can go with the `rstuf_auth` service PR? It would make more sense to make these changes in the context of the designed service when it's implemented (we also keep this PR small and remove unused fields - if they are indeed unused. I am waiting for someone else also to test this as well because for me it's working also for `token/new` :smiley:)

MVrachev:
Together with @KAUTH we examined the RSTUF API code again and we agreed on a couple of points:
1. currrently there is no way to create a new user in the RSTUF API database. You can only issue new tokens.
2. you cannot change the username or password of the current only existing user `admin` without changing the hardcoded values. There should be a way for a user to add new users or at least change his user name and password without touching the source code
3. Because of 1 and 2 this leads to a situation where you cannot issue a token to another user than `admin` with password `secret` because the check for a new token [here](https://github.com/vmware/repository-service-tuf-api/blob/3c1a3534795dba89802bfe8cff738912f80c0b49/repository_service_tuf_api/token.py#L159) will fail as in the database there is no user than `admin` with password `secret`.
4. it will be good to change the name of the `token_data` argument in `token.post()`: https://github.com/vmware/repository-service-tuf-api/blob/3c1a3534795dba89802bfe8cff738912f80c0b49/repository_service_tuf_api/token.py#L157 as it sounds like this is information coming from a token, but instead is information that comes from the request payload


I suggest we think of a name for this feature resolving steps from `1` to `3` or file issues.
Those problems doesn't seem to block us for the `Minimum Working Version`, but for sure are required for the `Minimum Viable Version`.
@kairoaraujo do you have any opinion on the matter?

kairoaraujo:
> Together with @KAUTH we examined the RSTUF API code again and we agreed on a couple of points:
> 
> 1. currrently there is no way to create a new user in the RSTUF API database. You can only issue new tokens.
> 2. you cannot change the username or password of the current only existing user `admin` without changing the hardcoded values. There should be a way for a user to add new users or at least change his user name and password without touching the source code
> 3. Because of 1 and 2 this leads to a situation where you cannot issue a token to another user than `admin` with password `secret` because the check for a new token [here](https://github.com/vmware/repository-service-tuf-api/blob/3c1a3534795dba89802bfe8cff738912f80c0b49/repository_service_tuf_api/token.py#L159) will fail as in the database there is no user than `admin` with password `secret`.
> 4. it will be good to change the name of the `token_data` argument in `token.post()`: https://github.com/vmware/repository-service-tuf-api/blob/3c1a3534795dba89802bfe8cff738912f80c0b49/repository_service_tuf_api/token.py#L157
>     as it sounds like this is information coming from a token, but instead is information that comes from the request payload
> 
> I suggest we think of a name for this feature resolving steps from `1` to `3` or file issues. Those problems doesn't seem to block us for the `Minimum Working Version`, but for sure are required for the `Minimum Viable Version`. @kairoaraujo do you have any opinion on the matter?

Ok. I will share some background and opinions here.

About items 1, 2, and 3.
The intention is not to create user management in this auth feature. We want to simplify it as much as possible as this is not a feature for RSTUF domain. For example, the idea is to work similarly to tokens in GitHub, where we don't manage users and passwords. We provide tokens with scopes from a main account. Once we go deep into this functionality, such as creating a user, resetting passwords, etc., it will be complex (two-factor auth .... 😄 ) .
The initial idea is to have a single administrator for RSTUF deployment. This administrator can generate tokens with scopes for other "users". Besides adding users, resetting passwords IMHO is more important to revoke tokens (#65 )

Suppose we decide to provide full authentication for RSTUF API. I suggest removing it from API, creating a new service (container), and adopting an existing service.

PR #265, opened 2023-02-20
Create an auth service for the rstuf api

kapsalis:
As it mentioned in
https://github.com/vmware/repository-service-tuf/issues/41 the API service has a simple built-in authentication and authorization implementation for small deployments.

We want to be able to disable this built-in feature and add support of other methods.

This commit is a first step on this direction. We move the authentication code to a new custom authentication service inside of the API service.

We achieve separation of concerns and we can easily replace the custom authentication service with another one without touching the business logic, if we have an authentication port with the same methods implemented for every different authentication type (e.g., custom service, aws api gateway).

This implementation helps us to avoid mocking, and replace the `CustomAuthenticationService` with a `TestAuthenticationService` when we are on the testing environment. The test service must implement the same methods, but don't need to access any db. It can store objects in memory.


kairoaraujo:
I'm adding some early review here, mainly to clarify some general aspects such as design and project direction.

First of all, the work is impressive and admirable. 

> As it mentioned in [vmware/repository-service-tuf#41](https://github.com/vmware/repository-service-tuf/issues/41) the API service has a simple built-in authentication and authorization implementation for small deployments.
> 
> We want to be able to disable this built-in feature and add support of other methods.
>
> This commit is a first step on this direction. We move the authentication code to a new custom authentication service inside of the API service.
> 
> We achieve separation of concerns and we can easily replace the custom authentication service with another one without touching the business logic, if we have an authentication port with the same methods implemented for every different authentication type (e.g., custom service, aws api gateway).

I like how you detached the authentication from the API, which can simplify disabling it.

A clarification here about [Disable option for built-in Authentication/Authorization](https://github.com/vmware/repository-service-tuf/issues/41)
We don't want to support other internal authentication methods. The idea is to have on/off for the built-in RSTUF Authentication. In principle, we don't want to allow the RSTUF user to replace it with a custom authentication service in RSTUF API. It can help you simplify, even more, your implementation.

We even thought of removing the built-in RSTUF Authorization/Authentication completely. In the end, we decided to keep a simple built-in RSTUF authentication only for organizations that don't have an auth infrastructure.

RSTUF users will be recommended to deploy/use their authentication by external services, such as API Gateways (SaaS/Container/Server) as a best practice.

```mermaid
C4Context
      title RSTUF API using external Authentication/Authorization
      Boundary(b0, "Public Environment") {
        Person(system_integration, "CI/CD or Distribution Platform", "Automated Integration")
        Person(rstuf_admin, "RSTUF Admin")
        Person_Ext(rstuf_user, "Developer", "PEP 480")

        System(api_gateway, "API Gateway", "API Gateway SaaS | Container | Server API Gateway")

        Enterprise_Boundary(b1, "Organization Environment") {
          System_Boundary(b2, "RSTUF Deployment") {
            System(rstuf_api_admin, "RSTUF API", "/api/v1/bootstrap | /api/v1/config | /api/v1/key_rotation")
            System(rstuf_api_general, "RSTUF API", "/api/v1/targets | /api/v1/tasks")
          }

          Boundary(b3, "Organization Auth", "Auth/RBAC") {
            System(auth, "User Auth Service")
          }
        }
      }

      Rel(system_integration, api_gateway, "Uses")
      Rel(rstuf_admin, api_gateway, "Uses")
      Rel(rstuf_user, api_gateway, "Uses")
      Rel(api_gateway, rstuf_api_admin, "")
      Rel(api_gateway, rstuf_api_general, "")
      Rel(api_gateway, auth, "")


      UpdateElementStyle(system_integration, $fontColor="red", $bgColor="grey", $borderColor="red")
      UpdateRelStyle(system_integration, api_gateway, $textColor="blue", $lineColor="blue", $offsetX="5")
      UpdateRelStyle(rstuf_admin, api_gateway, $textColor="blue", $lineColor="blue", $offsetX="5")
      UpdateRelStyle(rstuf_user, api_gateway, $textColor="blue", $lineColor="blue", $offsetX="5")
      UpdateRelStyle(api_gateway, auth, $textColor="blue", $lineColor="red", $offsetX="1")

      UpdateLayoutConfig($c4ShapeInRow="2", $c4BoundaryInRow="2")
```

kapsalis:
Hi @kairoaraujo . Thank you for the detailed reply. When an external service is used, how does rstuf_api authorizes the user? I was thinking that the scopes should remain part of the authorization, so a token must be sent in any case. However, the service does not expose any endpoints to issue a token.

If this is not the case, we can do the following:

 - Have an env variable that indicates if we use the built-in authentication/authorization service
 - In this [function](https://github.com/vmware/repository-service-tuf-api/pull/265/files#diff-e83760af1f30605c87f15abf3a13c265ec3430d59eec94c7007214bda6810908R113) (it can also be a decorator and use it in the endpoints instead of using the fastapi one), that calls the built-in authentication service to validate the token, we skip the check when the variable indicates that we don't want the built-in authentication service
 - We hide the token endpoints

kairoaraujo:
> Hi @kairoaraujo . Thank you for the detailed reply. When an external service is used, how does rstuf_api authorizes the user?

When the built-in Auth is disabled in RSTUF API, it no longer authorizes the user anymore.
The RSTUF API becomes a Rest API without authentication/authorization.

The user must define their new authentication and authorization model based on their rules (iScopes, Tokens, RBAC etc).
This is defined when creating the API Gateway, examples (server [ngnix](https://www.nginx.com/blog/deploying-nginx-plus-as-an-api-gateway-part-1/#implement-auth), [saas aws gateway](https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-control-access-to-api.html))

>I was thinking that the scopes should remain part of the authorization, so a token must be sent in any case. However, the service does not expose any endpoints to issue a token.

We cannot guarantee that our Scope policy is what the user needs for the access control, so we prefer to move everything to the API Gateways level.

> If this is not the case, we can do the following:
> 
> * Have an env variable that indicates if we use the built-in authentication/authorization service

👍 
IMHO, it should be disabled by default.

> * In this [function](https://github.com/vmware/repository-service-tuf-api/pull/265/files#diff-e83760af1f30605c87f15abf3a13c265ec3430d59eec94c7007214bda6810908R113) (it can also be a decorator and use it in the endpoints instead of using the fastapi one), that calls the built-in authentication service to validate the token, we skip the check when the variable indicates that we don't want the built-in authentication service

That sounds like a good approach. Having it as a decorator makes it easy to maintain.

> * We hide the token endpoints

Currently, we have a [documented environment variable](https://github.com/vmware/repository-service-tuf-api/blob/main/docs/source/guide/Docker_README.md#optional-rstuf_tokens_node) that disables the token endpoints, for scenarios where the RSTUF Admin want to deploy public APIs with limited access (without bootstrap, tokens endpoints) for example.
Having the RSTUF API Auth disabled, could automatically [set it](https://github.com/vmware/repository-service-tuf-api/blob/89c97c5cf501b381a7c4775142e991d314410c4b/app.py#L34) to `False`.

kapsalis:
Very nice! Thank you for the details! I will work on this changes on top of this PR the next days.

kairoaraujo:
I will update the original issue with the details we added here.
Thank you @kapsalis, this is an amazing contribution to RSTUF

kairoaraujo:
Also, will it be appropriate to update the following documentation. I can help if you need me.

- devel/design
https://github.com/vmware/repository-service-tuf-api/blob/0a2d7dea2d6a4503761982faf61ec975ca886fe4/docs/source/devel/design.rst?plain=1#L14

- guide/api
https://github.com/vmware/repository-service-tuf-api/blob/0a2d7dea2d6a4503761982faf61ec975ca886fe4/docs/source/guide/api.rst?plain=1#L5

- diagram/C3 level
https://github.com/vmware/repository-service-tuf-api/blob/0a2d7dea2d6a4503761982faf61ec975ca886fe4/docs/diagrams/repository-service-tuf-api-C3.puml#L1

kapsalis:
Remaining work:

 - Add UTs for the Repositories
 - Add extra UTs to the auth service (e.g., test expired token, or issue token with invalid expired_delta)
 - Create a FakeAuthService to use on API testing (use dynaconf environment)
 - Check which token API UTs are not needed, because some tests cases may be tested in service level UTs
 - Documentation requested by @kairoaraujo
 
 This MR start becoming huge. When the draft is ready, I can split it into more MRs (e.g., add the service first, then apply the changes to the API level). @kairoaraujo  let me know if you want to proceed this way.

tests/unit/rstuf_auth/services/test_auth.py
```diff
@@ -0,0 +1,107 @@
...
+        )
+
+        self.auth_service = CustomAuthenticationService(
+            secrets_settings=self.mocked_secret_settings,
+            user_repo=self.user_repo,
+            scope_repo=self.scope_repo,
+            user_scope_repo=self.user_scope_repo,
+        )
+
+    def test_initiate_admin(self):
```
kapsalis, line 40: Add a comment that admin has been created at line 33 (in class constructor)

kairoaraujo:
> Remaining work:
> 
> * Add UTs for the Repositories
> * Add extra UTs to the auth service (e.g., test expired token, or issue token with invalid expired_delta)
> * Create a FakeAuthService to use on API testing (use dynaconf environment)
> * Check which token API UTs are not needed, because some tests cases may be tested in service level UTs
> * Documentation requested by @kairoaraujo
> 
> This MR start becoming huge. When the draft is ready, I can split it into more MRs (e.g., add the service first, then apply the changes to the API level). @kairoaraujo let me know if you want to proceed this way.

I think we should move this discussion to the issue itself.
We might create a plan to merge this feature.

kairoaraujo:
Closing this PR as we removed the Authentication/Authorization feature.
- #472 
- https://github.com/repository-service-tuf/repository-service-tuf/issues/491

PR #269, opened 2023-02-22
Flexible maintainers CD Workflow

kairoaraujo:
This PR introduces a change in the `cd.yml` Workflow.

We release our components versions using the tags `v*`
The process of release has two steps:
1. Run the functional tests (FT) using the component to be released `v*` tag (from `main` branch) against the latest releases of the components.
2. If the (FT) finishes successfully, the `cd.yml` releases the new Docker Image GitHub Registry and adds it to our Releases page in Github. (Python package in case of the CLI).

If step 1 fails, the release breaks the compatibility with the current release.

The problem we are solving is how to release a new version when it breaks the compatibility.

This PR adds the workflow dispatch option for releases.
The maintainers can now go to the Actions > CD > Run Workflow, choose the `v*` tag release, and specify the version of the components it is compatible with, including `dev` (main branch) to run the FT tests. 

Using the Workflow dispatcher process, the Maintainers should bump to a version that shows it is breaking compatibility and make it clear on the release page.
We have a [Release Process](https://repository-service-tuf.readthedocs.io/en/latest/devel/release.html#release-process) and we are not changing it, only adding the Workflow that allows those actions.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -1,22 +1,63 @@
 name: CD
 
 on:
-  workflow_dispatch:
   push:
     tags:
-      - v*
+      - v*    
+  workflow_dispatch:      
+    inputs:
+      api_version:
+        description: "API Version"
+        default: "dev"
+        type: string
+        required: False
+      worker_version:
+        description: "Worker Version"
+        default: "latest"
+        type: string
+        required: False
+      cli_version:
+        description: "CLI Version"
+        default: "latest"
+        type: string
+        required: False
+
 
 jobs:
-  functional-latest:
+  set-component-versions:
+    runs-on: ubuntu-latest
+    outputs:
+      api_version: ${{ steps.api_version.outputs.VERSION }}
+      worker_version: ${{ steps.worker_version.outputs.VERSION }}
+      cli_version: ${{ steps.cli_version.outputs.VERSION }}
+
+    steps:
+      - id: api_version
+        name: dynamic input api version
+        run: |
+          if [[ "${{ inputs.api_version }}" == "" ]]; then echo "VERSION=dev" >> $GITHUB_OUTPUT; else echo "VERSION=${{ inputs.api_version }}" >> $GITHUB_OUTPUT;fi
+
+      - id: worker_version
+        name: dynamic input worker version
+        run: |
+          if [[ "${{ inputs.worker_version }}" == "" ]]; then echo "VERSION=latest" >> $GITHUB_OUTPUT; else echo "VERSION=${{ inputs.worker_version }}" >> $GITHUB_OUTPUT;fi
+
+      - id: cli_version
+        name: dynamic input cli version
+        run: |
+          if [[ "${{ inputs.cli_version }}" == "" ]]; then echo "VERSION=latest" >> $GITHUB_OUTPUT; else echo "VERSION=${{ inputs.cli_version }}" >> $GITHUB_OUTPUT;fi
+
+  functional-tests:
+    needs: set-component-versions
     uses: vmware/repository-service-tuf/.github/workflows/functional.yml@main
     with:
-      worker_version: latest
-      api_version: dev
-      cli_version: latest
+      api_version: ${{ needs.set-component-versions.outputs.api_version }}   
+      worker_version: ${{ needs.set-component-versions.outputs.worker_version }}   
+      cli_version: ${{ needs.set-component-versions.outputs.cli_version }}   
 
   release:
     runs-on: ubuntu-latest
-    needs: functional-latest
+    needs: functional-tests
     steps:
     - name: Checkout release tag
       uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -65,9 +65,9 @@ jobs:
     - name: Add final tags
       run: |
         docker pull ghcr.io/vmware/repository-service-tuf-api:${{ github.sha }}
-        docker tag ghcr.io/vmware/repository-service-tuf-api:${{ github.sha }} ghcr.io/vmware/repository-service-tuf-api:latest
         docker tag ghcr.io/vmware/repository-service-tuf-api:${{ github.sha }} ghcr.io/vmware/repository-service-tuf-api:${{ github.ref_name }}
         docker push ghcr.io/vmware/repository-service-tuf-api:${{ github.ref_name }}
+        docker tag ghcr.io/vmware/repository-service-tuf-api:${{ github.sha }} ghcr.io/vmware/repository-service-tuf-api:latest
         docker push ghcr.io/vmware/repository-service-tuf-api:latest
 
     - name: Publish GitHub Release
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -3,14 +3,9 @@ name: CD
 on:
   push:
     tags:
-      - v*    
-  workflow_dispatch:      
+      - v*
+  workflow_dispatch:
     inputs:
-      api_version:
-        description: "API Version"
-        default: "dev"
-        type: string
-        required: False
       worker_version:
         description: "Worker Version"
         default: "latest"
@@ -24,7 +19,13 @@ on:
 
 
 jobs:
+  publish-container-image:
+    uses: vmware/repository-service-tuf/.github/workflows/publish_container.yml@kairoaraujo-flexible-cd
+    with:
+      image_version: ${{ github.ref_name }}
+
   set-component-versions:
+    needs: publish-container-image
     runs-on: ubuntu-latest
     outputs:
       api_version: ${{ steps.api_version.outputs.VERSION }}
@@ -32,11 +33,6 @@ jobs:
       cli_version: ${{ steps.cli_version.outputs.VERSION }}
 
     steps:
-      - id: api_version
-        name: dynamic input api version
-        run: |
-          if [[ "${{ inputs.api_version }}" == "" ]]; then echo "VERSION=dev" >> $GITHUB_OUTPUT; else echo "VERSION=${{ inputs.api_version }}" >> $GITHUB_OUTPUT;fi
-
       - id: worker_version
         name: dynamic input worker version
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/cd.yml
```diff
@@ -22,28 +22,39 @@ on:
...
+      run: |
+        if [[ ${{ inputs.worker_version }} == "" ]]; then echo dev; else echo ${{ inputs.worker_version }}; fi
+
+    - id: cli_version
+      name: dynamic input cli version
       run: |
-        if [[ ${{ inputs.api_version }} == "" ]]; then echo "API_VERSION=${{ inputs.api_version }}" >> $GITHUB_ENV; fi
-        if [[ ${{ inputs.worker_version }}  == "" ]]; then echo "WORKER_VERSION=${{ inputs.worker_version }}" >> $GITHUB_ENV; fi
-        if [[ ${{ inputs.cli_version }}  == "" ]]; then echo "CLI_VERSION=${{ inputs.cli_version }}" >> $GITHUB_ENV; fi
+        if [[ ${{ inputs.cli_version }} == "" ]]; then echo dev; else echo ${{ inputs.cli_version }}; fi
```
MVrachev: I am wondering if `inputs.api_version` is not provided are we sure it will be equal to `""`?
Are we sure it will exist at all and there won't be an error when you try to access an attribute that doesn't exist?

.github/workflows/cd.yml
```diff
@@ -22,28 +22,39 @@ on:
...
+      worker_version: ${{ steps.worker_version.outputs.result }}
+      cli_version: ${{ steps.cli_version.outputs.result }}
+
     steps:
-    - name: use custom if workflow_dispatch
+
+    - id: api_version
+      name: dynamic input api version
+      run: |
+        if [[ ${{ inputs.api_version }} == "" ]]; then echo dev; else echo ${{ inputs.api_version }}; fi
```
MVrachev: Are you sure you shouldn't `echo "dev"`? 
I ask because you expect `inputs.api_version` to be a string and are we sure that `dev` only won't be interpreted as something else and not a string?

.github/workflows/cd.yml
```diff
@@ -32,21 +32,20 @@ jobs:
...
-        if [[ "${{ inputs.worker_version}}" == "" ]]; then echo dev; else echo ${{ inputs.worker_version }}; fi
-
-    - id: cli_version
-      name: dynamic input cli version
-      run: |
-        if [[ "${{ inputs.cli_version }}" == "" ]]; then echo dev; else echo ${{ inputs.cli_version }}; fi
+      - id: api_version
+        name: dynamic input api version
+        run: |
+          if [[ "${{ inputs.api_version }}" == "" ]]; then echo dev >> $GITHUB_OUTPUT; else echo ${{ inputs.api_version }} >> $GITHUB_OUTPUT; fi
```
MVrachev: Why do we save the `inputs.api_version` in the `GITHUB_OUTPUT` env variable?

MVrachev (commented):
Thanks for spending the time researching and fixing this problem!
It's never fun dealing with CI and infrastructure, but you spend a good amount of work on this.

It was little confusing to review this pr as it had multiple commits that could be squashed into one. A clear example of that are commits: https://github.com/vmware/repository-service-tuf-api/pull/269/commits/c9fc19bfd58a8675df1b4b72b9b9b10cbd03ee4f, https://github.com/vmware/repository-service-tuf-api/pull/269/commits/47eae6dfaa58081961f76679f5bddb58c1ffb990 and https://github.com/vmware/repository-service-tuf-api/pull/269/commits/b447afff7b0392d74cc899c4c8bdab00616eef0c which can be squashed/ammended into one.

When I see a reasonable amount of commits I like to review them one by one in order to get an understanding of the flow of thoughts that lead to the final change.
A couple of times I left a pending comment on one of your commits just to see that the thing I was asking for was removed in the next commit.

.github/workflows/cd.yml
```diff
@@ -1,18 +1,51 @@
...
+      cli_version:
+        description: "CLI Version"
+        default: "latest"
+        type: string
+        required: False      
+  workflow_call:
+    inputs:
+      api_version:
+        description: "API Version"
+        default: "dev"
```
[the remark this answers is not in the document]
kairoaraujo: Yes, that is correct for both questions.

But I'm still not happy with it in relation to using the `dev`

The `dev` will be the latest merged code from `main` and will be submitted to the GitHub Container Registry (GHCR) as Docker Image (We do it on the `publish_docker_dev.yml` in every main merge.
So, the FT will use something which is not the `v*` or commit chosen in the Workflow Dispatcher.

The correct way to go is when we trigger the CD:
1.  Build a container image with the chosen `v*` tag, but using commit id as the version
2. Publish the image to the GHC
3. Trigger the FT using this image version (commit)
4. If the tests pass generate the tag using the `v*` tag and the `latest`
5. Publish the release

.github/workflows/cd.yml
kairoaraujo: Yes, it is correct. I've tested here.
https://github.com/vmware/repository-service-tuf-api/actions/runs/4242147999/jobs/7373275082

.github/workflows/cd.yml
kairoaraujo: I think it is not the latest commit.

.github/workflows/cd.yml
kairoaraujo: This is not the latest commit.
Best hide this comment.

.github/workflows/cd.yml
kairoaraujo: This is not the latest commit.
Best hide this comment.

kairoaraujo:
> It was little confusing to review this pr as it had multiple commits that could be squashed into one. A clear example of that are commits: [c9fc19b](https://github.com/vmware/repository-service-tuf-api/commit/c9fc19bfd58a8675df1b4b72b9b9b10cbd03ee4f), [47eae6d](https://github.com/vmware/repository-service-tuf-api/commit/47eae6dfaa58081961f76679f5bddb58c1ffb990) and [b447aff](https://github.com/vmware/repository-service-tuf-api/commit/b447afff7b0392d74cc899c4c8bdab00616eef0c) which can be squashed/ammended into one.

> When I see a reasonable amount of commits I like to review them one by one in order to get an understanding of the flow of thoughts that lead to the final change. A couple of times I left a pending comment on one of your commits just to see that the thing I was asking for was removed in the next commit.

Yes, I can squashe them. This is because you need the commits to be able to test, re-test, do again as the GH Actions there is not a playground and real test is a required :/

kairoaraujo:
I will move this PR to Draft, I will need to test[ it again applying what I mentioned above.](https://github.com/vmware/repository-service-tuf-api/pull/269#discussion_r1115750425)

.github/workflows/publish_container.yml
```diff
@@ -0,0 +1,54 @@
...
+      image_version:
+        description: "Image version"
+        default: "dev"
+        type: string
+        required: False
+  workflow_call:
+    inputs:
+      image_version:
+        description: "Image version"
+        default: "dev"
```
MVrachev, line 15: Why the default should be `dev`?
Shouldn't it be a reference to the tag used when calling this `yml` file or in other words: `${{ github.ref_name }}`

.github/workflows/publish_container.yml
```diff
@@ -0,0 +1,54 @@
...
+        type: string
+        required: False
+
+jobs:
+  publish_container_image:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c
```
MVrachev, line 24: In the name of consistency and clarity maybe it will be better to add:
`- name: Checkout release tag` as it was in the `cd.yml` workflow.

.github/workflows/cd.yml
```diff
@@ -64,8 +64,10 @@ jobs:
 
     - name: Add final tags
       run: |
-        docker pull ghcr.io/vmware/repository-service-tuf-api:${{ github.ref_name }}
-        docker tag ghcr.io/vmware/repository-service-tuf-api:${{ github.ref_name }} latest
+        docker pull ghcr.io/vmware/repository-service-tuf-api:${{ github.sha }}
```
MVrachev, line 67: Why did you need to pull the image from the GitHub container registry?
Don't you have to build it locally?

.github/workflows/cd.yml
```diff
@@ -64,8 +64,10 @@ jobs:
...
+        docker tag ghcr.io/vmware/repository-service-tuf-api:${{ github.sha }} ghcr.io/vmware/repository-service-tuf-api:latest
+        docker tag ghcr.io/vmware/repository-service-tuf-api:${{ github.sha }} ghcr.io/vmware/repository-service-tuf-api:${{ github.ref_name }}
```
MVrachev, line 69: Why do you need to tag the same image as `latest` and with its `sha`?
Given that you execute the commands one after another do you there is a possibility someone will override `latest` and it won't point out to the same image?
Maybe it's good to do it that way from a security perspective.

MVrachev (commented):
I think it's better.
I have a few questions to clarify and make sure I got what you did here, but I hope we can merge it really soon.

.github/workflows/cd.yml
kairoaraujo: It was build in another job by another workflow.

.github/workflows/cd.yml
kairoaraujo: > Why do you need to tag the same image as `latest` and with its `sha`?
I don't tag the same image as `latest` and `sha`. I tag the `sha` with the additional tag `latest`

The `docker tag` command is `docker tag <image:tag> <image:aditional-tag>`

> Given that you execute the commands one after another do you there is a possibility someone will override `latest` and it won't point out to the same image?
> Maybe it's good to do it that way from a security perspective.

I'm reordering the commands to release first the `github.ref_name`

The latest will be added only for CD workflows. The Maintainers should handle a race condition. The process of releasing shouldn't be done without a PR approved before and alignment between the component maintainers. (we still not having the license to add the approval step in GH Actions)

Another option is to eliminate the `latest` tag for our Docker Images.

.github/workflows/publish_container.yml
kairoaraujo: The default is dev, once the most common case will be publishing the main (our `dev` tag) for merged commits.

MVrachev: But isn't `${{ github.ref_name }}` the same as `dev`? 
I read it's the commit sha of the commit starting the workflow.
If there are equal probably `dev` is clearer, but just wanted to make sure I got this correct.

.github/workflows/cd.yml
MVrachev: If you mean in the `publish_container.,yml` workflow I want to point out that there the default `image_version` is `dev` and not `github.sha` meaning we are using `dev` and `github.sha` together.
If I understand the github documentation well enough they should be the same which is fine.

.github/workflows/cd.yml
MVrachev: I got what you did.
I agree we are the ones who start the release process and if we are diligent the probability of race condition is close to zero.

MVrachev (approved):
I have only one comment left, other than that approving.
EDIT: we discussed it and everything is clear now.

Issue #183, opened 2022-11-16
Bug: UT fails updating python dependencies (starlette)


kairoaraujo:
### What happened?

https://github.com/vmware/repository-service-tuf-api/actions/runs/3477471009/jobs/5813668802

The above error caused Starlette to update to 0.21.0.
For more details, see:
- https://github.com/tiangolo/fastapi/issues/5646
- https://github.com/encode/starlette/pull/1953

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

kairoaraujo:
We can remove the pinned version when the fix for Starlette is merged and there is a new release.

kairoaraujo:
Closed by
- #270

PR #270, opened 2023-02-23
remove the pinned starlette version

kairoaraujo:
It removes the pinned starlette dependency version

This pinned version was applied as a workaround for issue #183

The startelle version is updated to the latest version, which fixes a security issue https://github.com/advisories/GHSA-74m5-2c7w-9w3x

The solution requires to install the new `httpx` development dependency to support the unit tests using the FastAPI TestClient as described in this link: https://fastapi.tiangolo.com/tutorial/testing/

A workaround needs to be applied to the tests which uses `DELETE` method.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -14,7 +14,6 @@ python-jose = "*"
 sqlalchemy = "*"
 redis = "*"
 bcrypt = "*"
-starlette = "==0.20.4"
 
 [dev-packages]
 black = "*"
@@ -40,6 +39,7 @@ mistune = "==0.8.4"
 myst-parser = "*"
 pre-commit = "*"
 bandit = "*"
+httpx = "*"
 
 [requires]
 python_version = "3.10"
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -59,10 +59,7 @@ pytest==7.2.1
 pytz==2022.7.1
 pyyaml==6.0 ; python_version >= '3.6'
 requests==2.28.2 ; python_version >= '3.7' and python_version < '4'
-<<<<<<< HEAD
-=======
 rfc3986[idna2008]==1.5.0
->>>>>>> c536857 (remove the pinned starlette version)
 setuptools==67.4.0 ; python_version >= '3.7'
 six==1.16.0 ; python_version >= '2.7' and python_version not in '3.0, 3.1, 3.2, 3.3'
 smmap==5.0.0 ; python_version >= '3.6'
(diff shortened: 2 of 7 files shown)


MVrachev (dismissed):

MVrachev:
Could you fix the conflict?

Pipfile
```diff
@@ -40,6 +39,7 @@ mistune = "==0.8.4"
 myst-parser = "*"
 pre-commit = "*"
 bandit = "*"
+httpx = "*"
```
KAUTH, line 42: Nit (to remind us in the future why we need it as a dev-package):

```suggestion
# Required to use `TestClient` from `fastapi`
httpx
```

requirements-dev.txt
```diff
@@ -55,9 +59,14 @@ pytest==7.2.1
 pytz==2022.7.1
 pyyaml==6.0 ; python_version >= '3.6'
 requests==2.28.2 ; python_version >= '3.7' and python_version < '4'
+<<<<<<< HEAD
```
KAUTH, line 62: As @MVrachev spotted there's a conflict and the tests are failing from the conflict markers left here.
(quick fix running `make requirements`)
After that, I run the tests locally and everything passes!

KAUTH (changes requested):
Besides the necessary change it looks good to me

Pipfile
kairoaraujo: I didn't add it because it fails straightforwardly in the TestClient import.

MVrachev: I agree it will be useful to add this comment.
I don't insist on this as well.

KAUTH: > I didn't add it because it fails straightforwardly in the TestClient import.

In that case, I see your point.

> It sounds like we should comment why we use all other dependencies rofl

Hahaha no that is not necessary :laughing: Jokes aside, I was mentioning this because I remember an issue that we had some time ago where I had to check the `Pipfile` requirements one by one because extra requirements had been accidentally added. So basically what I did was to check which ones were being imported in our source code (`from <package-name> import x` or `import <package-name>`). In this case I would have probably accidentally removed `httpx` from the `Pipfile`.
But indeed, if you run the tests as you mentioned it would have probably be an easy fix to see that `httpx` would be missing.

KAUTH (approved):

Pipfile
MVrachev: Great argumentation @KAUTH.

MVrachev (approved):

PR #274, opened 2023-02-28
Release v0.0.2a1

kairoaraujo:
Component Interoperability Matrix

* RSTUF Worker >= v0.0.2a1
* RSTUF CLI >= v0.0.1a9

### DIFF
--- a/repository_service_tuf_api/__version__.py
+++ b/repository_service_tuf_api/__version__.py
@@ -4,6 +4,6 @@
 #
 # SPDX-License-Identifier: MIT
 
-version = "0.0.1a4"
+version = "0.0.2a1"
 copyright = "Copyright (c) 2022-2023 VMware Inc"
 author = "Kairo de Araujo"


repository_service_tuf_api/__version__.py
```diff
@@ -4,6 +4,6 @@
 #
 # SPDX-License-Identifier: MIT
 
-version = "0.0.1a4"
+version = "0.0.2a1"
```
MVrachev, line 7: You are going to add the `v` in front of the version after I merge it or?

kairoaraujo: We use the `v` for the tag in the git.
https://github.com/vmware/repository-service-tuf-api/blob/f74de8613955ad25b44df6a27c39a2fe27cbcd1d/.github/workflows/cd.yml#L4-L6

MVrachev: And because you will start the `cd.yml` workflow manually that's why you won't name starting with `v` correct?

kairoaraujo: There is no reason, to be honest.
It is because we always kept in this file the semantic version without the `v` flag.

PR #286, opened 2023-03-08
Fix: Change flake8 exclude config

### LINKED ISSUE #237: Bug: tox lint commands reports warnings
### What happened?

The output of the lint checks, i.e. https://github.com/vmware/repository-service-tuf-api/pull/235/checks#step:5:65 , includes several warnings about the use of a glob matching pattern.

### What steps did you take?

_No response_

Only warnings or errors introduced by a PR are reported in the checks output.

### Relevant log output

```shell
> Warning:  The 'exclude' field in hook 'flake8' is a regex, not a glob -- matching '/*' probably isn't what you want here
```

### Code of Conduct

KAUTH:
- Removed the "exclude" settings of `flake8`
from `tox.ini` because they were overiding the
ones in `.pre-commit-config.yaml`

- Redefined the files/dirs to be excluded for
`flake8` in `.pre-commit-config.yaml`. Specifically:
        - took into account that the patterns for `exclude` are
	python regular expressions and are matched with `re.search`,
	and not globals as for `tox`.
	(see https://pre-commit.com/#regular-expressions)
        - `pre-commit` won’t ever pass untracked files to `flake8`
	so excluding files that are in `.gitignore` is unecessary
	(e.g., `.venv`, `venv`, `.tox`, etc.).
	See https://flake8.pycqa.org/en/latest/user/using-hooks.html#usage-with-the-pre-commit-git-hooks-framework

Closes #237


.pre-commit-config.yaml
```diff
@@ -13,7 +13,7 @@ repos:
   rev: '6.0.0'
   hooks:
   - id: flake8
-    exclude: repository_service_tuf_api/__init__.py|venv|.venv|setting.py|.git|.tox|dist|docs|/*lib/python*|/*egg|build|tools
+    exclude: docs
```
MVrachev, line 16: Removing `repository_service_tuf_api/__init__.py` from the `exclude` list means we would lint `repository_service_tuf_api/__init__.py` with `flake8` from now on as `repository_service_tuf_api/__init__.py` is not ignored in `.gitignore`.
That's what we want, correct?

MVrachev (dismissed):
I have only one question, otherwise LGTM!

.pre-commit-config.yaml
MVrachev: It seems to be a legit python file. I don't see a reason why we won't lint it.

MVrachev (approved):

PR #301, opened 2023-03-28
Simplifies metadata bootstrap process

### LINKED ISSUE #228: Simplify the API payload contract design
### What is the task about?

Related to the bigger task https://github.com/vmware/repository-service-tuf/issues/23 we have decided to simplify the bootstrap process https://github.com/vmware/repository-service-tuf/issues/208.

We should remove `timestamp`, `snapshot`, `timestamp`, and `bins` roles from the API payload expected body.

### Parent feature

https://github.com/vmware/repository-service-tuf/issues/208

### Code of Conduct

enyinna1234:
This removes the string parameter from the metadata and replaces it with root and target roles in the BootstrapPayload.

Closes #228


tests/data_examples/bootstrap/payload.json
```diff
@@ -106,130 +106,6 @@
...
-        "spec_version": "1.0.30",
-        "expires": "2023-01-14T20:23:45Z",
-        "meta": {
-          "snapshot.json": {
-            "version": 2
-          }
-        }
-      }
-    },
     "1.root": {
```
kairoaraujo, line 109:
```suggestion
    "root": {
```

repository_service_tuf_api/bootstrap.py
```diff
@@ -152,7 +152,13 @@ class Settings(BaseModel):
 
 class BootstrapPayload(BaseModel):
     settings: Settings
-    metadata: Dict[str, TUFMetadata]
+    metadata: Dict[
+        Literal[
+            Roles.ROOT.value,
+            Roles.TARGETS.value,
```
kairoaraujo, line 158: I suggest we keep only `Roles.ROOT.value`

```suggestion
            Roles.ROOT.value,
```

tests/data_examples/bootstrap/payload.json
kairoaraujo, line 1: We can also remove the `1.targets` entry.

kairoaraujo (changes requested):

kairoaraujo:
Thank you @enyinna1234 for your PR 🎉 

- To fix the lint is a simple: https://github.com/vmware/repository-service-tuf-api/pull/301/checks#step:5:83 (You can also run locally `tox -e lint`)
- The failure test is more about the `1.root`, and it expects `root`.

MVrachev:
Also, @enyinna1234 before submitting a pr for an issue please assign yourself or if you don't have the permissions, leave a comment on the issue.

MVrachev:
Maybe as part of this pr we can enforce that the metadata we are expecting from the bootstrap process is actually exactly only `root`.
This is a decision we discussed with @kairoaraujo and which he suggested here https://github.com/vmware/repository-service-tuf-cli/pull/232#discussion_r1147793756 and I implemented in my latest changes in commit: https://github.com/vmware/repository-service-tuf-cli/pull/232/commits/20add326c17b6b9e5de979805f060e04cb9376da

tox.ini
```diff
@@ -44,7 +44,7 @@ allowlist_externals =
     rm
     plantuml
 commands =
-    plantuml -Djava.awt.headless=true -o ../source/_static/ -tpng docs/diagrams/*.puml
+    # plantuml -Djava.awt.headless=true -o ../source/_static/ -tpng docs/diagrams/*.puml
```
MVrachev, line 47: Can you remove this change?
I have done the same before to comment `plantuml` when it's failing but we don't want it to be commented on by default otherwise it will never be used.

enyinna1234: I see. I'm trying to fix the issue of test-docs failing in the PR that's why that was commented out as a workaround. But it didn't work.

repository_service_tuf_api/bootstrap.py
```diff
@@ -113,7 +113,7 @@ class Settings(BaseModel):
 
 class BootstrapPayload(BaseModel):
     settings: Settings
-    metadata: Dict[str, TUFMetadata]
```
MVrachev, line 116: @kairoaraujo we are changing the `BootstrapPayload` expected values.
Shouldn't there be a unit test failure?

kairoaraujo:
> you can change the `root` name from `payload[metadata]` to `1.root` for example. This way we can make sure that if the payload contains information that's not expected we would indeed catch it. You can call `test_client.post()` with `with pytest.raises` see this example: https://github.com/vmware/repository-service-tuf-cli/blob/2f6e9bf42d086e093b7817f727f8d3910a7779fc/tests/unit/cli/admin/test_ceremony.py#L132 where you will have to check what exception will be raised.
> 
> @kairoaraujo am I right with my proposal for this unit test?

Yes, you're right. 👍 
Comment:  it will not raise, it will give a `reponse.status_code` as 422 if you're using `test_client`.

The simplest one IMO is just to create a fake payload like done in this test
https://github.com/vmware/repository-service-tuf-api/blob/caee05b94dfa82b9c78ffc4ac4551c9ac9c699c0/tests/unit/api/test_bootstrap.py#L159

Example
```python
payload = {"settings": {}, "metadata": {"timestamp": {}}
```
You can assert the status code as 422
```python
assert response.status_code == status.HTTP_422_UNPROCESSABLE_ENTITY
```

MVrachev (approved):

PR #311, opened 2023-04-07
Add a switch to remove the built-in authentication

### LINKED ISSUE #282: Task: Implement a simple on/off and make tokens & authentication/authorization
### What is the task about?

The goal is a simple implementation to disable Tokens and Authentication/Authorization (auth) disabled in the REST API.

Once the container RSTUF API has the environment variable `RSTUF_AUTH=false`, it disables the feature for Authentication/Authorization.

This feature doesn't need to be refactored or elaborated, it will be done in a specific task.

### Parent feature

- https://github.com/vmware/repository-service-tuf/issues/41

### References

_No response_

### Code of Conduct

MVrachev:
Introduce `RSTUF_AUTH` env variable which will allow for external
authentication providers to be plugged in.

Also, exclude all token-related API
calls when authentication is not used.

For the rest of the API calls introduce a switch that allows
to disable their authentication.

Close #282

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -94,6 +94,8 @@ services:
       - RSTUF_REDIS_SERVER="redis://redis"
       - SECRETS_RSTUF_TOKEN_KEY="secret"
       - SECRETS_RSTUF_ADMIN_PASSWORD="secret"
+      # If you want to test without authentication uncomment the next line:
+      # - RSTUF_AUTH=false
     volumes:
       - ./:/opt/repository-service-tuf-api:z
     depends_on:
--- a/docs/source/guide/Docker_README.md
+++ b/docs/source/guide/Docker_README.md
@@ -87,11 +87,16 @@ These settings are shared with the repository workers
 
 Important: It should use the same db id as used by RSTUF Workers.
 
-#### (Required) `SECRETS_RSTUF_TOKEN_KEY`
+#### (Required) `RSTUF_AUTH`
+Disable/Enable RSTUF built-in token authentication. Default: true
+
+Note: Disable this feature only if you have set up another authentication provider.
+
+##### (Required) `SECRETS_RSTUF_TOKEN_KEY`
 
 Secret Token for hash the Tokens.
 
-#### (Required) `SECRETS_RSTUF_ADMIN_PASSWORD`
+##### (Required) `SECRETS_RSTUF_ADMIN_PASSWORD`
 
 Secret admin password.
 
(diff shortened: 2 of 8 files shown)


MVrachev:
From my research I have found that `auth` should be a lambda otherwise Fast API requires `args` and `kwargs` parameters for the API calls:
Untitled

Using lambda function which returns `None` seems to be correct way as it hides the `Authentication` button and doesn't show `args` and `kwargs` fields:
Untitled2

MVrachev:
This was tested manually and requires integration tests (with an RSTUF instance similar to the functional tests) to properly test it automatically.
There is an issue with this #312.

MVrachev:
I decided to use `RSTUF_AUTH` by default because of fear of any of our users who can deploy RSTUF without adding any authentication.

docs/source/guide/Docker_README.md
```diff
@@ -87,11 +87,16 @@ These settings are shared with the repository workers
 
 Important: It should use the same db id as used by RSTUF Workers.
 
-#### (Required) `SECRETS_RSTUF_TOKEN_KEY`
+#### (Required) `RSTUF_AUTH`
+Disable/Enable RSTUF built-in token authentication. Default: true
+
+Note: Disable this feature only if you have set up another authentication provider.
```
kairoaraujo, line 93: IMO we don't need this disclaimer/note. It will depend a lot on the architecture deployment. Maybe the API will be an internal service only, not exposed for use (might be a case for PyPI, for example)

repository_service_tuf_api/api/__init__.py
```diff
@@ -1,3 +1,24 @@
...
 #
 # SPDX-License-Identifier: MIT
+
+import logging
+from typing import Callable
+
+from repository_service_tuf_api import settings
+from repository_service_tuf_api.token import validate_token
+
+logging.basicConfig(
```
kairoaraujo, line 11: We don't need this declaration again. It is done in the 'app.py'

docker-compose.yml
```diff
@@ -94,6 +94,8 @@ services:
       - RSTUF_REDIS_SERVER="redis://redis"
       - SECRETS_RSTUF_TOKEN_KEY="secret"
       - SECRETS_RSTUF_ADMIN_PASSWORD="secret"
+      # If you want to test without authentication uncomment the next line:
```
KAUTH, line 97: Hey just passing by - I believe we don't need this comment and just want a `- RSTUF_AUTH=true` line, which should be the default as described in the docs

repository_service_tuf_api/api/__init__.py
```diff
@@ -1,3 +1,24 @@
...
+    level=logging.DEBUG,
+    format="%(asctime)s %(levelname)s %(message)s",
+    datefmt="%H:%M:%S",
+)
+
+
+def get_auth() -> Callable:
+    if settings.get("AUTH", True) is True:
```
KAUTH, line 19: `if settings.get("RSTUF_AUTH", True) is True:` ?

PR #323, opened 2023-04-18
Feature: add endpoint  /api/v1/metadata for metadata/key rotation

### LINKED ISSUE #289: Task: Implement new endpoint `PUT /api/v1/metadata`
### What is the task about?

This task is a continuation of #288, and it requires the task completed

- [x] #288 

It is about implementing the support for the RSTUF API for Key Rotation.

In this task we will implement a new endpoint `PUT /api/v1/metadata`.
This endpoint will receive a payload that contains metadata (as the method `POST`)

The implementation of this endpoint will submit the received and validated payload as a task  `key_rotation` to the broker.

### Parent feature

- https://github.com/vmware/repository-service-tuf/issues/23

### References

_No response_

### Code of Conduct

kairoaraujo:
Add `/api/v1/metadata` with method `POST` for Metadata/key rotation.

This feature allows RSTUF administrators to rotate the root metadata
and the online keys, which require signing with offline keys (root
keys).

This endpoint publishes a message/task in the RSTUF configured
message queue. The RSTUF Worker consumes the queue and executes the
metadata rotation.

The implementation is done in 3 steps:
### 1. refactoring for `metadata.py`.

This python package name conflicts with the new `/api/v1/metadata`
naming.

The current implementation on `repository_service_tuf_api/metadata.py`
package are core initialization required across other implementations
and were moved to `repository_service_tuf_api/__init__.py`.

It doesn't affect or change the user usage of the RSTUF API.

### 2. makes the TUF BaseModels reusable in a common package.

These models will be used to implement `/api/v1/metadata`

### 3. Feature: add endpoint  `/api/v1/metadata` for metadata/key rotation

Depends on: https://github.com/vmware/repository-service-tuf-cli/issues/222

Closes #289


tests/unit/api/test_metadata.py
```diff
@@ -0,0 +1,189 @@
...
+
+        with open(
+            "tests/data_examples/metadata/update-root-payload.json"
+        ) as f:
+            f_data = f.read()
+
+        payload = json.loads(f_data)
+        response = test_client.post(url, json=payload, headers=token_headers)
+
+        assert response.status_code == status.HTTP_200_OK
```
MVrachev, line 68: Looking into the status code I am wondering:
shouldn't we return non 200 status code when bootstrap doesn't exist?
I am asking because the `POST` operation cannot be completed successfully.
The same question is for the REST API calls where we return status code 200 when bootstrap is `False` and it should be `True`.

tests/unit/api/test_metadata.py
```diff
@@ -0,0 +1,189 @@
...
+        mocked_repository_metadata = pretend.stub(
+            apply_async=pretend.call_recorder(lambda *a, **kw: None)
+        )
+        monkeypatch.setattr(
+            "repository_service_tuf_api.metadata.repository_metadata",
+            mocked_repository_metadata,
+        )
+        monkeypatch.setattr(
+            "repository_service_tuf_api.metadata.get_task_id", lambda: "123"
+        )
```
MVrachev, line 111: Do we need all of that when the token is invalid?
Wouldn't this fail when arguments are evaluated?

tests/unit/api/test_metadata.py
```diff
@@ -0,0 +1,189 @@
...
```
MVrachev, line 157: Same question as above: do we need this when I think the failure should happen when evaluating the `Security` argument?

MVrachev (changes requested):

tests/unit/api/test_metadata.py
kairoaraujo: Good catch

tests/unit/api/test_metadata.py
kairoaraujo: Good catch

repository_service_tuf_api/metadata.py
```diff
@@ -0,0 +1,87 @@
...
+        example = json.loads(content)
+        schema_extra = {"example": example}
+
+
+class PostData(BaseModel):
+    task_id: Optional[str]
+
+
+class MetadataPostResponse(BaseModel):
+    data: Optional[PostData]
```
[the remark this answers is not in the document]
kairoaraujo: A good example is the 500 status code with an internal server error handled by FastAPI.
We need to be clear to the API user that we can have a response without body/JSON data.

repository_service_tuf_api/metadata.py
```diff
@@ -0,0 +1,87 @@
...
+        with open(
+            "tests/data_examples/metadata/update-root-payload.json"
+        ) as f:
+            content = f.read()
+        example = json.loads(content)
+        schema_extra = {"example": example}
+
+
+class PostData(BaseModel):
+    task_id: Optional[str]
```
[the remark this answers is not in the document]
kairoaraujo: You can fail during generating the `task_id`, and it will return status code 500 (if a internal server error) without body, for example.

MVrachev:
I am going to approve it, but I think you need to update the docs again.

MVrachev (approved):

PR #334, opened 2023-04-25
More flexible disabling endpoint feature

kairoaraujo:
This feature gives better flexibility to the RSTUF user to disable endpoints.

The user can disable specific endpoints and methods in a simple configuration using the `RSTUF_DISABLE_ENDPOINTS`.

It reduces the number of settings parameters (environment variables)` RSTUF_BOOTSTRAP_NODE`, and `RSTUF_TOKEN_NODE`, making the feature more flexible as we add more endpoints and methods to the RSTUF API.

### DIFF
--- a/app.py
+++ b/app.py
@@ -51,13 +51,15 @@ def load_endpoints():
         disabled_endpoints.append(f"{api_v1.prefix}{token_v1.prefix}/")
 
     for v1_endpoint in v1_endpoints:
-        for endpoint_route in v1_endpoint.routes:
+        endpoint_routes = v1_endpoint.routes.copy()
+        for endpoint_route in endpoint_routes:
             route = (
                 f"{endpoint_route.methods}{api_v1.prefix}{endpoint_route.path}"
             )
             if route in disabled_endpoints:
                 logging.info(f"Disabled endpoint {route}")
                 v1_endpoint.routes.remove(endpoint_route)
+
         if f"{api_v1.prefix}{v1_endpoint.prefix}/" in disabled_endpoints:
             logging.info(
                 f"Disabled endpoint {api_v1.prefix}{v1_endpoint.prefix}/"
--- a/app.py
+++ b/app.py
@@ -41,26 +41,34 @@ api_v1 = APIRouter(
     responses={404: {"description": "Not found"}},
 )
 
-# load endpoints
-disabled_endpoints: List[str] = settings.get("DISABLE_ENDPOINTS", "").split(
-    ":"
-)
-if is_auth_enabled is False:
-    disabled_endpoints.append(f"{api_v1.prefix}{token_v1.prefix}/")
 
-for v1_endpoint in v1_endpoints:
-    for endpoint_route in v1_endpoint.routes:
-        route = f"{endpoint_route.methods}{api_v1.prefix}{endpoint_route.path}"
-        if route in disabled_endpoints:
-            logging.info(f"Disabled endpoint {route}")
-            v1_endpoint.routes.remove(endpoint_route)
-    if f"{api_v1.prefix}{v1_endpoint.prefix}/" in disabled_endpoints:
-        logging.info(f"Disabled endpoint {api_v1.prefix}{v1_endpoint.prefix}/")
-    else:
-        api_v1.include_router(v1_endpoint)
+def load_endpoints():
+    # load endpoints
+    disabled_endpoints: List[str] = settings.get(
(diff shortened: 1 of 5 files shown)


MVrachev:
Can you also update the docs?

kairoaraujo:
@MVrachev it is ready now

tests/unit/test_app.py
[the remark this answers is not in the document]
kairoaraujo: Thank you for the suggestion. I applied it :)

docs/source/guide/Docker_README.md
```diff
@@ -111,6 +93,29 @@ to `/run/secrets` path.
...
+
+Note: It will be disabled all methods and all paths GET `/api/v1/targets`,
+POST `/api/v1/targets`, POST `/api/v1/targets/publish` etc.
+
+It is possible to disable specific method endpoint given
+`{'POST'}/api/v1/targets/publish`.
+
+Note: If you give both
+`RSTUF_DISABLE_ENDPOINTS={'POST'}/api/v1/targets/publish:/api/v1/targets` the
+`/api/v1/targets` has higher weigh and will disable all
```
MVrachev, line 113:
```suggestion
`RSTUF_DISABLE_ENDPOINTS={'POST'}/api/v1/targets/publish:/api/v1/targets` then
the `/api/v1/targets` has a higher priority and will disable all v1 targets related endpoints.
```

docs/source/guide/Docker_README.md
```diff
@@ -111,6 +93,29 @@ to `/run/secrets` path.
...
+
+A list can be given as example bellow
```
MVrachev, line 115:
```suggestion
A list can be given as shown in the example bellow:
```

app.py
```diff
@@ -27,24 +38,35 @@
...
+            route = (
+                f"{endpoint_route.methods}{api_v1.prefix}{endpoint_route.path}"
+            )
+            if route in disabled_endpoints:
+                logging.info(f"Disabled endpoint {route}")
+                v1_endpoint.routes.remove(endpoint_route)
+        if f"{api_v1.prefix}{v1_endpoint.prefix}/" in disabled_endpoints:
+            logging.info(
+                f"Disabled endpoint {api_v1.prefix}{v1_endpoint.prefix}/"
+            )
```
MVrachev, line 61: In your test case, you disable:
`"{'POST'}/api/v1/targets/publish/:/api/v1/targets/"`
I wonder wouldn't the logger log that first you disable `{'POST'}/api/v1/targets/publish/:` and 
then log that you disable the whole `/api/v1/targets/"` endpoint?
At least that seems to me to be the case when I look at the code.

app.py
```diff
@@ -15,6 +16,16 @@
...
 
+v1_endpoints = [
+    bootstrap_v1,
+    config_v1,
+    metadata_v1,
+    targets_v1,
+    tasks_v1,
+    token_v1,
+]
+is_authentication_enabled: bool = settings.get("AUTH", True)
```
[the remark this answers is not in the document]
kairoaraujo: That is correct! I found that a rebase was needed. (not sure why GitHub didn't complain)
Thank you for catching it.

app.py
kairoaraujo: Yes, that is correct!
I fixed it. I noticed I was interacting and changing the list. Not a best practice.
Another great spot! 🎉

app.py
```diff
@@ -31,23 +41,37 @@
...
+            )
+        else:
+            api_v1.include_router(v1_endpoint)
 
-rstuf_app.include_router(api_v1)
+    rstuf_app.include_router(api_v1)
+
+
+load_endpoints()
+logging.info(f"Bootstrap ID: {settings_repository.get_fresh('BOOTSTRAP')}")
```
MVrachev, line 74: Are you sure there will be `bootstrap` when you import the `app.py`?

app.py
MVrachev: By changing the `endpoint_route` do you mean when you remove it from v1:
`v1_endpoint.routes.remove(endpoint_route)`
or I missed something :D

app.py
kairoaraujo: No, but then it will log that the bootstrap is None.
```logs
repository-service-tuf-api_1     | 10:40:59 INFO Bootstrap ID: None
```

app.py
kairoaraujo: Yes, we shouldn't interact and update the list at the same. It changes the indexes.

MVrachev (approved):

PR #335, opened 2023-04-25
Update container docs about secrets

kairoaraujo:
Add the mention about the secrets standard adopted by RSTUF. RSTUF requires user to mount the secrets volume in `/run/secrets` path.


MVrachev (commented):
This feature you document Is supported by default, right?
Without us doing anything?

MVrachev (approved):
From conversations with Kairo, I understood that we do parse `/run/secrets`.

PR #347, opened 2023-05-15
FT workflow requires secrets

kairoaraujo:
It is required after merging
- repository-service-tuf/repository-service-tuf#336


KAUTH (approved):
LGTM!

PR #353, opened 2023-05-23
Fix: protection during bootstrap process

### LINKED ISSUE #342: Bug: Improve protection during bootstrap process
### What happened?

Currently, we have the following process for bootstrap:
1. An API request sends the bootstrap task to the broker
2. The API registers the bootstrap as `pre-<task-id>` in Redis
3. The API returns the `task-id` to the user and starts monitoring the `task-id` status with a timeout  (default: 300 seconds)
4. A Worker picks up that task
5. The Worker makes the bootstrap process
6. The Worker finishes the bootstrap process and bootstrap register in Redis to `task-id` (removing the `pre-`)

During step 2, if the task stays longer than the timeout, it will be canceled ([Revoking](https://docs.celeryq.dev/en/stable/reference/celery.app.control.html#celery.app.control.Control.revoke) the task)

The problem with the bootstrap consistency between steps 2 and 6. If the user sends a request to add a target to the TUF Metadata, the API will identify that bootstrap is complete as `pre-<task-id>` is considered as **done**

### What steps did you take?

1. Using the RSTUF Worker source code
2. Add a long sleep to the bootstrap: `import time; time.sleep(299)`
3. Run the development environment for Worker `make run-dev`
4. Open the http://localhost
5. Call the `/api/v1/bootstrap`
6. Try to call it again, and you will receive that `System has already a metadata`
```json
{
  "detail": {
    "error": "System already has a Metadata."
  }
}
```
8. Try to add targets `/api/v1/targets`
9. You will notice that the task will be failed `/api/v1/tasks`
```json
{
  "data": {
    "task_id": "b76a41c95996423dbb34eca376196812",
    "state": "FAILURE",
    "result": "Can't open Role 'targets'"
  },
  "message": "Task state."
}
```
10. You can see the exception in the Worker
```log
repository-service-tuf-worker_1  |   File "/opt/repository-service-tuf-worker/repository_service_tuf_worker/services/storage/local.py", line 64, in get
repository-service-tuf-worker_1  |     raise StorageError(f"Can't open Role '{role}'") from e
repository-service-tuf-worker_1  | securesystemslib.exceptions.StorageError: Can't open Role 'targets'
```

The RSTUF API doesn't allow adding targets during `pre-` bootstrap

### Relevant log output

_No response_

### Code of Conduct

kairoaraujo:
This change considers the bootstrap `pre` lock done by the RSTUF API and avoid user can start adding/removing targets before the process finishes.

Closes #342

### DIFF
--- a/docs/swagger.json
+++ b/docs/swagger.json
@@ -810,6 +810,10 @@
                     "bootstrap": {
                         "title": "Bootstrap",
                         "type": "boolean"
+                    },
+                    "state": {
+                        "title": "State",
+                        "type": "string"
                     }
                 }
             },
--- a/repository_service_tuf_api/__init__.py
+++ b/repository_service_tuf_api/__init__.py
@@ -202,10 +202,7 @@ def is_bootstrap_done():
     # correctly reload it because that and the settings_repository.reload() do
     # the job.
     settings_repository.reload()
-    if settings_repository.get_fresh("BOOTSTRAP", None) is None:
-        return False
-    else:
-        return True
+    return settings_repository.get_fresh("BOOTSTRAP")
 
 
 def get_task_id():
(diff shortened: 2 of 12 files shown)


kairoaraujo:
I will submit a new approach, totally different from this PR.

PR #389, opened 2023-07-31
Fix: protection during bootstrap process

### LINKED ISSUE #342: Bug: Improve protection during bootstrap process

kairoaraujo:
The checks now are aligned with the Repository Settings `BOOTSTRAP` as mentioned in https://repository-service-tuf.readthedocs.io/en/stable/devel/design.html#rstuf-repository-settings-configuration

This change considers the bootstrap `pre` lock done by the RSTUF API and avoids the user can start adding/removing targets before the process finishes.

# Description

Fixes #342

# Types of changes
- [X] Bug fix (non-breaking change which fixes an issue)

# Additional requirements
- [X] Tests have been added for the bug fix or new feature
- [X] Docs have been added for the bug fix or new feature

# Code of Conduct

### DIFF
--- a/repository_service_tuf_api/__init__.py
+++ b/repository_service_tuf_api/__init__.py
@@ -4,6 +4,7 @@
 
 import logging
 import os
+from dataclasses import dataclass
 from enum import Enum
 from typing import Optional
 from uuid import uuid4
@@ -27,6 +28,13 @@ logging.basicConfig(
     format="%(asctime)s %(levelname)s %(message)s",
     datefmt="%H:%M:%S",
 )
+
+
+@dataclass
+class BootstrapState:
+    bootstrap: bool
+    state: Optional[str] = None
+    task_id: Optional[str] = None
 
 
 class SCOPES_NAMES(Enum):
@@ -196,20 +204,40 @@ def release_bootstrap_lock():
     redis_loader.write(settings_repository, settings_data)
 
 
-def is_bootstrap_done():
+def bootstrap_state() -> BootstrapState:
     """
-    Check if the boot is done.
+    Bootstrap state
+
+    The bootstrap state is registered in Redis.
+    Detailed definitions are available in
+    https://repository-service-tuf.readthedocs.io/en/stable/devel/design.html#tuf-repository-settings  # noqa
     """
+
     # Reload the settings
     # The reload is required because the settings object is created in the
     # `app.py`'s initialization. The `settings_repository.get_fresh() doesn't
     # correctly reload because of that the settings_repository.reload() do
     # the job.
     settings_repository.reload()
-    if settings_repository.get_fresh("BOOTSTRAP", None) is None:
-        return False
-    else:
-        return True
+    bootstrap = settings_repository.get_fresh("BOOTSTRAP")
+
+    bootstrap_state = BootstrapState(bootstrap=False, state=None, task_id=None)
+    if bootstrap is None:
+        return bootstrap_state
+
+    if len(bootstrap.split("-")) == 1:
+        bootstrap_state.bootstrap = True
+        bootstrap_state.state = "finished"
+        bootstrap_state.task_id = bootstrap
+
+        return bootstrap_state
+
+    elif len(bootstrap.split("-")) == 2:
+        bootstrap_state.bootstrap = False
+        bootstrap_state.state = bootstrap.split("-")[0]
+        bootstrap_state.task_id = bootstrap.split("-")[1]
+
+        return bootstrap_state
 
 
 def get_task_id():
--- a/repository_service_tuf_api/__init__.py
+++ b/repository_service_tuf_api/__init__.py
@@ -226,6 +226,7 @@ def bootstrap_state() -> BootstrapState:
         return bootstrap_state
 
     if len(bootstrap.split("-")) == 1:
+        # This is a finished bootstrap. It only contains the `<task-id>``
         bootstrap_state.bootstrap = True
         bootstrap_state.state = "finished"
         bootstrap_state.task_id = bootstrap
@@ -233,6 +234,8 @@ def bootstrap_state() -> BootstrapState:
         return bootstrap_state
 
     elif len(bootstrap.split("-")) == 2:
+        # This is considered an intermediated state It is not finished because
+        # there is a `<state>-`
         bootstrap_state.bootstrap = False
         bootstrap_state.state = bootstrap.split("-")[0]
         bootstrap_state.task_id = bootstrap.split("-")[1]
--- a/tests/unit/api/test_bootstrap.py
+++ b/tests/unit/api/test_bootstrap.py
@@ -13,10 +13,12 @@ class TestGetBootstrap:
         self, test_client, token_headers, monkeypatch
     ):
         url = "/api/v1/bootstrap/"
-        mocked_check_metadata = pretend.call_recorder(lambda: False)
+        mocked_bootstrap_state = pretend.call_recorder(
+            lambda *a: pretend.stub(bootstrap=False, state=None, task_id=None)
+        )
         monkeypatch.setattr(
-            "repository_service_tuf_api.bootstrap.is_bootstrap_done",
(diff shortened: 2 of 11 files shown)


repository_service_tuf_api/__init__.py
```diff
@@ -194,20 +202,40 @@ def release_bootstrap_lock():
...
+        return bootstrap_state
+
+    if len(bootstrap.split("-")) == 1:
+        bootstrap_state.bootstrap = True
+        bootstrap_state.state = "finished"
+        bootstrap_state.task_id = bootstrap
+
+        return bootstrap_state
+
+    elif len(bootstrap.split("-")) == 2:
```
MVrachev, line 233: I think it will be easier to understand if you check if the `bootstrap` value starts with `pre`.
I just realized it's possible that a real task id could start with `pre` prefix.
Then I see two possibilities:
1. instead of using `pre-<task_id>` as an intermediate state why don't you use capital letters e.g. `PRE-<task_id>`? 
I see we don't use capital letters for task ids.
2. If you don't like the first solution, then at least leave a comment on what will be the structure of the `bootstrap` value.

tests/unit/api/test_bootstrap.py
```diff
@@ -76,16 +76,6 @@ def test_get_bootstrap_incorrect_scope_token(
             "Authorization": f"Bearer {token.json()['access_token']}",
         }
         url = "/api/v1/bootstrap/"
-        mocked_check_metadata = pretend.call_recorder(lambda: False)
```
MVrachev, line 79: We still check the bootstrap state.
Why can we remove this part of the code?

repository_service_tuf_api/__init__.py
```diff
@@ -29,6 +30,13 @@
 )
 
 
+@dataclass
+class BootstrapState:
+    bootstrap: bool
+    state: Optional[str] = None
```
MVrachev, line 36: I think it does make sense to make the `state` as an Enum, so that it's clear which are the allowed values.
What do you think?

tests/unit/api/test_bootstrap.py
```diff
@@ -243,23 +299,6 @@ def test_post_bootstrap_invalid_token(self, test_client, monkeypatch):
 
         token_headers = {"Authorization": "Bearer h4ck3r"}
 
-        mocked_check_metadata = pretend.call_recorder(lambda: False)
```
MVrachev, line 246: We can remove those mocks as we catch the empty payload in `bootstrap_state`, but before we didn't or what?

repository_service_tuf_api/__init__.py
kairoaraujo: I don't think you can have "pre-"
We use the https://github.com/repository-service-tuf/repository-service-tuf-api/blob/129f4dcadbc0bb0f2e4ba63f0112bd42172e3c8c/repository_service_tuf_api/__init__.py#L213-L214 which gives us a hexadecimal string without dashes (`-`).

tests/unit/api/test_bootstrap.py
kairoaraujo: It will fail on the API level (fastapi) because the token is invalid.
It is not processed.

repository_service_tuf_api/__init__.py
kairoaraujo: I thought to have it. I even did it, and I reverted it.
I reverted because it will be inconsistent if we add more "`<state>-`" in the future.
If you see it now, we don't make decisions on the `state` level. We only use the `bootstrap`.

The state will be used in specific decisions to sign the bootstrap, which requires "`signing-`".
Isn't it better to leave the control to the logic in that case?

repository_service_tuf_api/__init__.py
MVrachev: Okay, you are right, then check for `pre-` or `signing-`. Those were the possible intermediate values, right?

kairoaraujo: Any intermediate `<state>-` is a `bootstrap=False`
We consider only `bootstrap=True`, when doesn't have intermediate states.
The intermediate states can be used by specific checks (at the moment, no use case but with DAS we will have one)

MVrachev:
Can you add the new `bootstrap_state` check for the new API endpoint `PUT /api/config`?

kairoaraujo:
Rebased

repository_service_tuf_api/__init__.py
MVrachev: Yes, that's my point. To make it clear.
If you don't want to make a difference to the check, then leave a comment that in this situation the bootstrap is at an intermediate state and that all intermediate state are in the form of `<STATE>-<task_id>.

repository_service_tuf_api/__init__.py
MVrachev: > The state will be used in specific decisions to sign the bootstrap, which requires "`signing-`".
> Isn't it better to leave the control to the logic in that case?

We will still leave the control to the logic of the signing case and we won't use it.
We will just make it easier for the next implementation to not use a hardcoded `signing` string, but instead the Enum values.

kairoaraujo: We can decide on it in the DAS implementation. There is no use case of `state` for this specific PR.

tests/unit/tuf_repository_service_api/test__init__.py
```diff
@@ -56,3 +38,33 @@ def test_release_bootstrap_lock(self):
...
+        repository_service_tuf_api.settings_repository = pretend.stub(
+            reload=pretend.call_recorder(lambda: None),
+            get_fresh=pretend.call_recorder(lambda *a: None),
+        )
+        result = repository_service_tuf_api.bootstrap_state()
+        assert result == repository_service_tuf_api.BootstrapState(
+            False, None, None
+        )
+
+    def test_get_bootstrap_state_signing(self):
```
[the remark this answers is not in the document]
kairoaraujo: Done!

PR #395, opened 2023-08-02
Add a simple alias router to /api/v1/artifacts

kairoaraujo:
This was a request/feedback received.

As the `targets` might sound confusing on the documentation and TUF specification, the term `artifacts` reduces the confusion between the target files and the target roles.

Also, in the future, when there is a delegation targets metadata feature, it can turn more confusing.


MVrachev (commented):
Nice idea.
We need to do that in more places and in the other components as well.
Do we have an issue about this?

kairoaraujo:
> Nice idea.
> We need to do that in more places and in the other components as well.
> Do we have an issue about this?

We can file an issue to update our documentation.
The good thing is that we are just appending the alias for the endpoint and not removing it.
We can update the docs in parallell.

MVrachev (approved):

PR #452, opened 2023-09-24
Remove invalid required: true in parameter schema.

simi:
Fixes problem with invalid specification. Not tested yet.

image

I have changed the code and regenerated the `swagger.json` file.

Part of #455 

# Types of changes


kairoaraujo:
Rebased

simi:
Looks good!

app.py
```diff
@@ -7,34 +7,63 @@
...
+        version=__version__.version,
+        openapi_version=OPENAPI_VERSION,
+        description=DESCRITPTION,
+        routes=rstuf_app.routes,
+    )
+    rstuf_app.openapi_schema = openapi_schema
+    return rstuf_app.openapi_schema
+
+
+rstuf_app.openapi = _custom_openapi
```
MVrachev, line 50: Shouldn't we call `_custom_openapi` here?
```suggestion
rstuf_app.openapi = _custom_openapi()
```

MVrachev (commented):
I need a reminder why we decided to pin a specific version of OPENAPI?
I remember there were warnings regarding our API methods, right?
Could you remind me what the problem was?

simi:
> I need a reminder why we decided to pin a specific version of OPENAPI?

If I remember well no features of 3.1 were used and 3.1 is not supported by common tools around yet.

app.py
kairoaraujo: No, it is correct.
https://fastapi.tiangolo.com/how-to/extending-openapi/#normal-fastapi

MVrachev (approved):

PR #499, opened 2023-12-01
Add trusted_root in `GET /api/v1/metadata/sign` response 

MVrachev:
# Description
We should return trusted_root in the GET /metadata/sign response to
support the case when:
1. bootstrap is completed
2. a new DAS root is published
3. a maintainer needs to sign this new root

To complete step 3. we need to have access to both
new root and latest trusted_root.

# Types of changes

# Additional requirements
- [ ] Tests have been added for the bug fix or new feature

# Code of Conduct

### DIFF
--- a/tests/unit/api/test_metadata.py
+++ b/tests/unit/api/test_metadata.py
@@ -229,8 +229,6 @@ class TestGetMetadataSign:
             elif setting == "TRUSTED_ROOT":
                 return fake_trusted_metadata
 
-            return None
-
         mocked_settings_repository = pretend.stub(
             reload=pretend.call_recorder(lambda: None),
             get=pretend.call_recorder(lambda a: get_role(a)),


repository_service_tuf_api/metadata.py
```diff
@@ -84,8 +84,13 @@ def post_metadata(payload: MetadataPostPayload) -> MetadataPostResponse:
     return MetadataPostResponse(data=data, message=message)
 
 
+class RolesData(BaseModel):
+    root: TUFMetadata
+    trusted_root: Union[TUFMetadata, Dict]  # Will be {} if root is bootstrap
```
kairoaraujo, line 89: Isn't it better to make it `Optional[TUFMetadata]`

repository_service_tuf_api/metadata.py
```diff
@@ -120,18 +125,31 @@ def get_metadata_sign() -> MetadataSignGetResponse:
...
+            md_response[role] = signing_role_dict
+
+            trusted_obj = settings_repository.get(f"TRUSTED_{role.upper()}")
+            md_response[f"trusted_{role}"] = {}
+            if trusted_obj is not None:
+                trusted_dict = trusted_obj.to_dict()
+                # If versions match, signing_role_obj is initial bootsrap root.
+                signing_ver = signing_role_dict["signed"]["version"]
+                if trusted_dict["signed"]["version"] != signing_ver:
+                    md_response[f"trusted_{role}"] = trusted_dict
```
kairoaraujo, line 143: I would leave it to the client API user.
They can apply their rules and handle them with more freedom.

kairoaraujo (changes requested):

repository_service_tuf_api/metadata.py
MVrachev: I did it like that as you suggested that we should always return `trusted_root` in the response.
I am okay to change that.

kairoaraujo (approved):

PR #503, opened 2023-12-14
Skip tests in coverage

KAUTH:
# Description

Configuration to skip tests in coverage.
Reference https://pytest-cov.readthedocs.io/en/latest/config.html#configuration.

# Types of changes

# Additional requirements
- [ ] Tests have been added for the bug fix or new feature

# Code of Conduct


MVrachev (commented):
Are you sure that you have achieved what you intended?
When I look the link on the coverage website from this pull requests I see the tests files again:
image


Additionally, I tried it locally and here is the result after running `tox -e test` and thus updating the information that coverage will use:
image


PS: It will be best if you leave a link to the coverage doc (above this `[run]` target) which mentions that the `[run]` target actually will be used from coverage to exclude a folder.
Honestly, it's really strange why they decided to use this way as for me it's not obvious that `[run]` will be used by coverage.

MVrachev:
What is even more confusing is that it seems that it's working on the CLI pr https://github.com/repository-service-tuf/repository-service-tuf-cli/pull/471
there on the coverage website, we see: 
image

KAUTH:
Yep, it didn't omit the tests because I forgot to push the `--cov-config=tox.ini` option to pytest, which tells it where to get the configuration from, it should work now.

> PS: It will be best if you leave a link to the coverage doc (above this [run] target) which mentions that the [run] target actually will be used from coverage to exclude a folder.
Honestly, it's really strange why they decided to use this way as for me it's not obvious that [run] will be used by coverage.

The naming is a bit strange indeed, I added the link to the commit message as I wanted to keep the configuration clean from comments, but I can add it since it's really not easy to understand (I was also considering it). Although, checking with git blame you can get that information also.

KAUTH:
> What is even more confusing is that it seems that it's working on the CLI pr [repository-service-tuf/repository-service-tuf-cli#471](https://github.com/repository-service-tuf/repository-service-tuf-cli/pull/471) there on the coverage website, we see: image

According to `coverage` docs "Only importable files (ones at the root of the tree, or in directories with an `__init__.py` file) will be considered." Source: https://coverage.readthedocs.io/en/6.5.0/source.html
We didn't have these `__init__.py` files, so I [added](https://github.com/repository-service-tuf/repository-service-tuf-cli/pull/471/commits/a37ed912c99c7c170a19990121c5ee4ff40e7388) them, because we want to explicitly ignore the tests in the settings of `coverage` and not just because they are not importable.

MVrachev (approved):

PR #513, opened 2024-01-05
Add support for `PUT /api/v1/metadata/online`

### LINKED ISSUE #399: Task: add a new PUT /api/v1/metadata endpoint forcing new metadata version
### What is the task about?

Add a new `PUT /api/v1/metadata` endpoint forcing new metadata version.
By using forcing a new metadata version the metadata expiration date will be changed to `datetime.now() + <ROLE NAME>_EXPIRATION> number of days` where  `<ROLE NAME>_EXPIRATION>` is a repository setting. See https://repository-service-tuf.readthedocs.io/en/latest/devel/design.html#tuf-repository-settings.

This endpoint will accept one or many of the online role names: `timestamp`, `snapshot` or `bins`.

This endpoint can be used in conjunction with the `PUT /api/v1/config` endpoint to update the expiration policies of one or more online roles and enforce that the new expiration changes are applied at the moment. Otherwise, the new expiration policies will be used when a role has expired.

In order to do so the user will do:
1. Call `PUT /api/v1/config` with a dictionary `X` with the online role(s) with their corresponding new expiration policy
2. Call the new `PUT /api/v1/metadata` endpoint passing the online roles you mentioned in `X`.

### Code of Conduct

MVrachev:
# Description

Add support for `PUT /api/v1/metadata` which will force a new metadata version
of a given set of online roles.

Fixes #399

# Types of changes

# Additional requirements

# Code of Conduct

### DIFF
--- a/docs/swagger.json
+++ b/docs/swagger.json
@@ -194,7 +194,7 @@
                 ],
                 "summary": "Force a new version of online metadata role(s).",
                 "description": "Force a new version of online metadata role(s). If the roles list is empty all roles will be updated. The new metadata version(s) will have extended expiration which will equal to: today + ROLE_NAME_EXPIRATION number of days, where ROLE_NAME_EXPIRATION is a tuf repository setting. Note: depending on which metadata role you want to update other online roles will likely be updated as well otherwise consistency will be lost.",
-                "operationId": "post_api_v1_metadata_online_post",
+                "operationId": "post_online_api_v1_metadata_online_post",
                 "requestBody": {
                     "content": {
                         "application/json": {
@@ -975,13 +975,7 @@
                 "properties": {
                     "roles": {
                         "items": {
-                            "type": "string",
-                            "enum": [
-                                "targets",
-                                "snapshot",
-                                "timestamp",
-                                "bins"
-                            ]
+                            "type": "string"
                         },
                         "type": "array",
                         "title": "Roles"
@@ -991,7 +985,13 @@
                 "required": [
                     "roles"
                 ],
-                "title": "MetadataOnlinePostPayload"
+                "title": "MetadataOnlinePostPayload",
+                "example": {
+                    "roles": [
+                        "targets",
+                        "snapshot"
+                    ]
+                }
             },
             "MetadataOnlinePostResponse": {
                 "properties": {
@@ -1015,7 +1015,14 @@
                     "data",
                     "message"
                 ],
-                "title": "MetadataOnlinePostResponse"
+                "title": "MetadataOnlinePostResponse",
+                "example": {
+                    "data": {
+                        "last_update": "2022-12-01T12:10:00.578086",
+                        "task_id": "7a634b556f784ae88785d36425f9a218"
+                    },
+                    "message": "Force new online metadata accepted."
+                }
             },
             "MetadataPostPayload": {
                 "properties": {
--- a/docs/swagger.json
+++ b/docs/swagger.json
@@ -143,6 +143,49 @@
             }
         },
         "/api/v1/metadata/": {
+            "put": {
+                "tags": [
+                    "Metadata"
+                ],
+                "summary": "Force a new version of online metadata role(s).",
+                "description": "Force a new version of online metadata role(s). The new metadata version(s) will have extended expiration which will equal to: today + ROLE_NAME_EXPIRATION number of days, where ROLE_NAME_EXPIRATION is a tuf repository setting. Note: depending on which metadata role you want to update other online roles will likely be updated as well otherwise consistency will be lost.",
+                "operationId": "put_api_v1_metadata__put",
+                "requestBody": {
+                    "content": {
+                        "application/json": {
+                            "schema": {
+                                "$ref": "#/components/schemas/MetadataPutPayload"
+                            }
+                        }
+                    },
+                    "required": true
+                },
+                "responses": {
+                    "202": {
+                        "description": "Successful Response",
+                        "content": {
+                            "application/json": {
+                                "schema": {
+                                    "$ref": "#/components/schemas/MetadataPutResponse"
+                                }
+                            }
+                        }
+                    },
+                    "404": {
+                        "description": "Not found"
+                    },
+                    "422": {
+                        "description": "Validation Error",
+                        "content": {
+                            "application/json": {
+                                "schema": {
+                                    "$ref": "#/components/schemas/HTTPValidationError"
+                                }
+                            }
+                        }
+                    }
+                }
+            },
             "post": {
                 "tags": [
                     "Metadata"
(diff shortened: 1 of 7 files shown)


MVrachev:
It's best to test this together with https://github.com/repository-service-tuf/repository-service-tuf-api/pull/513

MVrachev:
Before merging this pr, lets merge:  #514.

MVrachev:
@kairoaraujo it's ready for a review.

repository_service_tuf_api/api/metadata.py
```diff
@@ -31,6 +31,25 @@ def post(payload: metadata.MetadataPostPayload):
     return metadata.post_metadata(payload)
 
 
+@router.put(
+    "/",
```
kairoaraujo, line 35: I think it should be a specific endpoint "/api/v1/metadata/<name>"
Also, I would use the following behavior. 
No rolename = bump all online roles.

repository_service_tuf_api/metadata.py
```diff
@@ -85,6 +85,63 @@ def post_metadata(payload: MetadataPostPayload) -> MetadataPostResponse:
...
+        raise HTTPException(
+            status.HTTP_200_OK,
+            detail={
+                "message": "Task not accepted.",
+                "error": (
+                    f"Requires bootstrap finished. State: {bs_state.state}"
+                ),
+            },
+        )
+
```
kairoaraujo, line 125: We should verify that it is an online role. See my comment on https://github.com/repository-service-tuf/repository-service-tuf-worker/pull/435 for more detail.

repository_service_tuf_api/api/metadata.py
kairoaraujo, line 35: I think it should be a specific endpoint "/api/v1/metadata/"
Also, I would use the following behavior.
No rolename = bump all online roles.

kairoaraujo (changes requested):

repository_service_tuf_api/api/metadata.py
MVrachev: It is already using `api/v1/metadata`:
image

I will address the second comment though.

MVrachev:
@kairoaraujo I did as you suggested if payload is empty I made sure force update on all roles.
I also added a check in `Roles.online_roles` to whether `Targets` is an online role.
If it's not we don't return it in this list.

docs/swagger.json
```diff
@@ -143,6 +143,49 @@
             }
         },
         "/api/v1/metadata/": {
+            "put": {
+                "tags": [
+                    "Metadata"
+                ],
+                "summary": "Force a new version of online metadata role(s).",
+                "description": "Force a new version of online metadata role(s). If the roles list is empty all roles will be updated. The new metadata version(s) will have extended expiration which will equal to: today + ROLE_NAME_EXPIRATION number of days, where ROLE_NAME_EXPIRATION is a tuf repository setting. Note: depending on which metadata role you want to update other online roles will likely be updated as well otherwise consistency will be lost.",
```
ivanayov, line 151: What will happen if `datetime.now() + <ROLE_NAME_EXPIRATION number of days>` is less than the current expiration? I think we should either make it impossible or define it more strictly here (like extended or redused - practically it can also be the same, so we can just say "updated").

ivanayov: What I mean is that this description creates expectations that don't strictly match the functionality.

repository_service_tuf_api/api/metadata.py
```diff
@@ -31,6 +31,26 @@ def post(payload: metadata.MetadataPostPayload):
...
+    summary="Force a new version of online metadata role(s).",
+    description=(
+        "Force a new version of online metadata role(s). If the roles list is "
+        "empty all roles will be updated. The new metadata version(s) will "
+        "have extended expiration which will equal to: "
```
ivanayov, line 40: Same about "extended" here.

docs/swagger.json
MVrachev: The whole idea is to be able to redefine current expiration.
For context: the idea of this feature came to life after I implemented `PUT /api/v1/config`.
With that API endpoint, you can change the expiration policy for each role.
Still, this policy will be enforced when an online role expires and the automatic job to bump that role is executed.
See: https://github.com/repository-service-tuf/repository-service-tuf-worker/blob/486e98cdbc16dd41c62c0ec8b473e2f94312ee9c/app.py#L139

The idea behind this new endpoint is that after you have called `PUT /api/v1/config` you can call this new endpoint `PUT /api/v1/metadata` and request a new metadata version of part or all online metadata roles.

For the future I have thought about adding a flag `force_new_version` to the `PUT /api/v1/config`  endpoint which will do the same think as this new endpoint but for all roles.

repository_service_tuf_api/api/metadata.py
MVrachev: Check my comment to your questions above.

docs/swagger.json
MVrachev: As suggested by Ivana I added a commit that makes sure we cannot push negative expiration days.
Here is the commit: https://github.com/repository-service-tuf/repository-service-tuf-api/pull/513/commits/8a00f5bb795d3b73dc3313924c019f543302d2ef

MVrachev:
I had to remove my latest changes where I added more requirements to our attributes as there was an issue.
I have added a new issue about this: #520.

repository_service_tuf_api/api/metadata.py
kairoaraujo: Sorry, what I mean is to use a more specific endpoint `/api/v1/metadata/<choose a name>`
I see the `/api/v1/metadata` for more generic actions like Update Metadata.

As this description says, "Force a new version of online metadata role(s)." it should have a more specific endpoint
i.e.: `/api/v1/metadata/online`. Following our past changes, it is an asynchronous task that could be a POST.

repository_service_tuf_api/metadata.py
```diff
@@ -85,6 +85,79 @@ def post_metadata(payload: MetadataPostPayload) -> MetadataPostResponse:
     return MetadataPostResponse(data=data, message=message)
 
 
+class MetadataPutPayload(BaseModel):
+    roles: List[Roles.online_roles_values()]
```
MVrachev, line 89: We need to change this as we would support custom target delegations' role names.
This should be any string.

repository_service_tuf_api/config.py
```diff
@@ -63,6 +63,8 @@ class Config:
 
 class Settings(BaseModel):
     # Only online roles can be dict keys
+    # TODO: Add requirement for int value after fixing: https://github.com/repository-service-tuf/repository-service-tuf-api/issues/377  # noqa
```
MVrachev, line 66: I need to remove those comments as we have removed the FAST API issue.

repository_service_tuf_api/api/metadata.py
MVrachev: After a discussion we agreed it will be `POST /api/v1/metadata/online/`

MVrachev:
@kairoaraujo ready to test together with https://github.com/repository-service-tuf/repository-service-tuf-worker/pull/435.
Support for pushing a new version of custom target delegation is there.

kairoaraujo (approved):

PR #518, opened 2024-01-21
refactoring: update the task response data in the `/api/v1/tasks`

### LINKED ISSUE #515: Task: change `state` value if worker task result has a `status = False`
### What is the task about?

This is a request coming from RubyGems maintainers that it would be better and easier for them if they can understand that a task has failed or errored based on the returned `state` instead of parsing the result and checking the `status`.
This will also improve all future clients.

### Parent feature

_No response_

### References

_No response_

### Code of Conduct

fsavoia:
… a error, false status

# Description
This is a request coming from RubyGems maintainers that it would be better and easier for them
if they can understand that a task has failed or errored based on the returned state instead of parsing
the result and checking the status.
This will also improve all future clients.

The more detailed changes are:

The `data.state` now includes the `ERRORED` state, an error identified by RSTUF while processing a task. This error was identified and handled by RSTUF, but might require attention from the user. Details will be included in `data.result`.
Screenshot 2024-01-30 at 06 33 36

The `data.result` now is a `Dict[string, Any]`. Before, it was `string` or `Dict[string, Any]`. The string was used in case of exception, which now has the same consistency.
Screenshot 2024-01-30 at 06 38 17

Success tasks will follow as
Screenshot 2024-01-30 at 06 38 57

Fixes #515 

# Types of changes

# Additional requirements

# Code of Conduct

### DIFF
--- a/repository_service_tuf_api/tasks.py
+++ b/repository_service_tuf_api/tasks.py
@@ -128,7 +128,9 @@ def get(task_id: str) -> Response:
 
     if isinstance(task.result, Exception):
         task_result = str(task.result)
-    elif task_state == TaskState.SUCCESS and not task_result["status"]:
+    elif task_state == TaskState.SUCCESS and not task_result.get(
+        "status", False
+    ):
         task_state = TaskState.ERRORED
 
     return Response(
--- a/repository_service_tuf_api/tasks.py
+++ b/repository_service_tuf_api/tasks.py
@@ -23,6 +23,7 @@ class TaskState(str, enum.Enum):
     REJECTED = states.REJECTED
     RETRY = states.RETRY
     IGNORED = states.IGNORED
+    ERRORED = "ERRORED"
     RUNNING = "RUNNING"  # custom state used when a task is RUNNING in RSTUF
 
 
@@ -72,6 +73,7 @@ class TasksData(BaseModel):
             "`RUNNING`: Task is running.\n\n"
             "`SUCCESS`: Task succeeded.\n\n"
             "`FAILURE`: Task failed.\n\n"
+            "`ERRORED`: Task errored.\n\n"
             "`REVOKED`: Task revoked.\n\n"
             "`REJECTED`: Task was rejected (only used in events).\n\n"
         )
@@ -120,12 +122,16 @@ def get(task_id: str) -> Response:
         ``Response`` as BaseModel from pydantic
     """
     task = repository_metadata.AsyncResult(task_id)
+
+    task_state = task.state
+    task_result = task.result
+
     if isinstance(task.result, Exception):
         task_result = str(task.result)
-    else:
-        task_result = task.result
+    elif task_state == TaskState.SUCCESS and not task_result["status"]:
+        task_state = TaskState.ERRORED
 
     return Response(
-        data=TasksData(task_id=task_id, state=task.state, result=task_result),
+        data=TasksData(task_id=task_id, state=task_state, result=task_result),
         message="Task state.",
     )
--- a/repository_service_tuf_api/tasks.py
+++ b/repository_service_tuf_api/tasks.py
@@ -5,7 +5,7 @@
 
 import enum
 from datetime import datetime
-from typing import Any, Optional, Union
+from typing import Any, Dict, Optional, Union
 
 from celery import states
 from pydantic import BaseModel, Field
@@ -49,7 +49,9 @@ class TaskDetails(BaseModel):
             "If the task status result is `False` shows an error message"
         )
     )
-    any: Any = Field(description="Any releavant information from task")
+    details: Optional[Dict[str, Any]] = Field(
+        description="Any releavant information from task"
+    )
 
 
 class TaskResult(BaseModel):
@@ -73,7 +75,8 @@ class TasksData(BaseModel):
             "`RUNNING`: Task is running.\n\n"
             "`SUCCESS`: Task succeeded.\n\n"
             "`FAILURE`: Task failed.\n\n"
-            "`ERRORED`: Task errored.\n\n"
+            "`ERRORED`: Task errored. RSTUF identified an error while "
+            "processing the task.\n\n"
             "`REVOKED`: Task revoked.\n\n"
             "`REJECTED`: Task was rejected (only used in events).\n\n"
         )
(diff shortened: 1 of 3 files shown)


repository_service_tuf_api/tasks.py
```diff
@@ -120,12 +122,16 @@ def get(task_id: str) -> Response:
...
     task = repository_metadata.AsyncResult(task_id)
+
+    task_state = task.state
+    task_result = task.result
+
     if isinstance(task.result, Exception):
         task_result = str(task.result)
-    else:
-        task_result = task.result
+    elif task_state == TaskState.SUCCESS and not task_result["status"]:
```
kairoaraujo, line 131:
```suggestion
    elif task_state == TaskState.SUCCESS and not task_result.get("status", False):
```

fsavoia: Good @kairoaraujo, thank you for your suggestion. I'm changing now.

repository_service_tuf_api/tasks.py
```diff
@@ -120,12 +122,18 @@ def get(task_id: str) -> Response:
...
-    else:
-        task_result = task.result
+    elif task_state == TaskState.SUCCESS and not task_result.get(
+        "status", False
+    ):
+        task_state = TaskState.ERRORED
 
     return Response(
-        data=TasksData(task_id=task_id, state=task.state, result=task_result),
+        data=TasksData(task_id=task_id, state=task_state, result=task_result),
```
MVrachev, line 137: Shouldn't we also change the `task.state = task_state` as if it enters on line `131` then `task_state` will change, but `task.state` will remain `TASK.SUCCESS`?
This sounds that could be misleading.
@kairoaraujo any thoughts?

tests/unit/api/test_tasks.py
```diff
@@ -17,6 +17,8 @@ def test_get(self, test_client, monkeypatch):
                 "status": True,
                 "task": "add_targets",
                 "last_update": "2023-11-17T09:54:15.762882",
+                "message": None,
```
MVrachev, line 20: Why there is a new `message` field when there is another `message` attribute under `details`?
I am a little confused.

tests/unit/api/test_tasks.py
```diff
@@ -72,3 +74,105 @@ def test_get_result_is_exception(self, test_client, monkeypatch):
...
+
+        assert test_response.json() == {
+            "data": {
+                "task_id": "test_id",
+                "state": "ERRORED",
+                "result": {
+                    "status": False,
+                    "task": "sign_metadata",
+                    "last_update": "2023-11-17T09:54:15.762882",
+                    "message": "Signature Failed",
```
MVrachev, line 109: Why do we need to have a `message` attribute at all when the state is clearly `ERRORED`?

MVrachev (commented):
I need some clarifications and maybe @kairoaraujo could join in with some wisdom as well :D

tests/unit/api/test_tasks.py
kairoaraujo: This is related to the changes where we move the `error` and `message` out from `details` to the task result

It is correct message is required.  Error is optional, actually it is required when `status` is `False` (it is also another PR in the worker for that)
Here is an example:
 
 https://github.com/repository-service-tuf/repository-service-tuf-worker/blob/486e98cdbc16dd41c62c0ec8b473e2f94312ee9c/repository_service_tuf_worker/repository.py#L625-L632

MVrachev: Okay, let's leave it then.

repository_service_tuf_api/tasks.py
kairoaraujo: Not sure if I understood.
We store `task.state` in `task_state`, and we no longer use the original object. (We do the same for `task_result`)

tests/unit/api/test_tasks.py
kairoaraujo: Details can have `Dict[str, Any]`, which could allow an extra message field.
But for clarity in the test, I agree we should remove it.
@fsavoia, I think it is good to remove it or add something else.

kairoaraujo (approved):

kairoaraujo (changes requested):
Following changes from:
- https://github.com/repository-service-tuf/repository-service-tuf-worker/pull/439/
---
Rename: https://github.com/repository-service-tuf/repository-service-tuf-api/blob/4c08b1d477e7cee41df2ad22f0c4593ff34ea426/repository_service_tuf_api/tasks.py#L55 to 
```python
class TaskReponse(BaseModel)`
```
---
Update

https://github.com/repository-service-tuf/repository-service-tuf-api/blob/4c08b1d477e7cee41df2ad22f0c4593ff34ea426/repository_service_tuf_api/tasks.py#L61

to
```python
details: Optional[TaskDetails] = Field(description="Relevant result details")
```
---

remove 

https://github.com/repository-service-tuf/repository-service-tuf-api/blob/4c08b1d477e7cee41df2ad22f0c4593ff34ea426/repository_service_tuf_api/tasks.py#L46-L51

---

Update
https://github.com/repository-service-tuf/repository-service-tuf-api/blob/4c08b1d477e7cee41df2ad22f0c4593ff34ea426/repository_service_tuf_api/tasks.py#L81-L85

to 
```python
 result: str = Field( 
     description=("Task result message" ) 
 ) 
 response: Optional[TaskResponse] = Field( 
     description=("Task response when `SUCCESS` or `ERRORED`" ) 
 ) 
 ```
 
 ---
 
 Add a example that follows the change

https://github.com/repository-service-tuf/repository-service-tuf-api/blob/4c08b1d477e7cee41df2ad22f0c4593ff34ea426/repository_service_tuf_api/tasks.py#L93-L105

suggestion:

```python
{
  "data": {
    "task_id": "33e66671dcc84cdfa2535a1eb030104c",
    "state": "SUCCESS",
    "result": "message",
    "response": {
      "status": true,
      "task": "add_targets",
      "last_update": "2023-11-17T09:54:15.762882",
      "details": {
        "artifacts": ["package.tar.gz"]
      }
    }
  },
  "message": "Task state."
}
```

kairoaraujo (approved):
Excellent work @fsavoia

PR #526, opened 2024-01-31
Feature: support to unrecognized fields (Root)

kairoaraujo:
# Description
This commit adds the support to unrecognized fields to the Root.

The Root.signed.unrecognized_fields can store relevant information in the TUF metadata used by TUF clients or even RSTUF Worker.

It is added to the `TUFSigned`, allowing extra fields. The unrecognized fields must to have the format `x-<vendor>-<name>`

We add a pydantic `root_validator` to verify all extra fields if they follow the required format.

Part of https://github.com/repository-service-tuf/repository-service-tuf/issues/608

# Types of changes

# Additional requirements

# Code of Conduct

# Note:
This support works with RSTUF Worker which doesn't require changes.

### DIFF
--- a/repository_service_tuf_api/common_models.py
+++ b/repository_service_tuf_api/common_models.py
@@ -87,7 +87,7 @@ class TUFSigned(BaseModel):
     class Config:
         fields = {"type": "_type"}
         # allow extra unrecognized fields (but it will be validated)
-        # https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.extra
+        # https://docs.pydantic.dev/1.10/usage/model_config/#options
         extra = Extra.allow
 
     # Custom Validator for the extra fields (TUF unrecognized_fields)
--- a/repository_service_tuf_api/common_models.py
+++ b/repository_service_tuf_api/common_models.py
@@ -4,9 +4,9 @@
 # SPDX-License-Identifier: MIT
 
 from enum import Enum
-from typing import Dict, List, Literal, Optional
+from typing import Any, Dict, List, Literal, Optional
 
-from pydantic import BaseModel, Field
+from pydantic import BaseModel, ConfigDict, Extra, Field, root_validator
 
 
 class Roles(Enum):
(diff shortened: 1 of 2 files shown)


repository_service_tuf_api/common_models.py
```diff
@@ -86,6 +86,36 @@ class TUFSigned(BaseModel):
 
     class Config:
         fields = {"type": "_type"}
+        # https://docs.pydantic.dev/latest/api/config/#pydantic.config.ConfigDict.extra
+        model_config = ConfigDict(
```
lukpueh, line 90: I am new to Pydantic and haven't seen this cnfig-as-nested-class pattern in the Pydantic docs. What I've seen is that model config is usually written into the `self.model_config` attribute inherited from `BaseModel`. Out of curiosity can you tell me more about the pattern you are using?

repository_service_tuf_api/common_models.py
```diff
@@ -86,6 +86,36 @@ class TUFSigned(BaseModel):
...
+            extra="allow",
+        )
+        extra = (
+            Extra.allow
+        )  # allow extra unrecognized fields (but it will be validated)
```
lukpueh, line 95: Without context there seems to be a bit of a definition redundancy. Would you mind telling me, why you need
`Config.model_config.extra = "allow"` and `Config.extra = Extra.allow`? :)

repository_service_tuf_api/common_models.py
```diff
@@ -86,6 +86,36 @@ class TUFSigned(BaseModel):
...
+
+    # Custom Validator for the extra fields (TUF unrecognized_fields)
+    @root_validator(pre=True)
+    def validate_unrecognized_fields(
+        cls, values: Dict[str, Any]
+    ) -> Dict[str, Any]:
+        all_required_field_names = {
+            field.alias
+            for field in cls.__fields__.values()
+            if field.alias != "extra"  # to support alias
```
lukpueh, line 105: Again, sorry for my ignorance, but where is the "extra" alias defined?

repository_service_tuf_api/common_models.py
```diff
@@ -86,6 +86,36 @@ class TUFSigned(BaseModel):
...
+            if field_name not in all_required_field_names:
+                if (
+                    not field_name.startswith("x")
+                    or len(field_name.split("-")) < 3
+                ):
+                    raise ValueError(
+                        f"Invalid: `{field_name}` field name, "
+                        "unrecognized_field must use format x-<vendor>-<name>"
+                    )
+        return values
```
lukpueh, line 118: Seems about right. 👍

repository_service_tuf_api/common_models.py
kairoaraujo: I'm not sure that I can answer you why I use it that way, but at least it is consistent in all the models for RSTUF API 😆 

Once upon a time,  Kairo was looking at how to create a field alias for the TUF _type field, and he found https://docs.pydantic.dev/1.10/usage/model_config/, which has an example.

repository_service_tuf_api/common_models.py
kairoaraujo: Sorry, my bad. Fixing.
It should have only `extra=Extra.allow`

repository_service_tuf_api/common_models.py
kairoaraujo: Oops

repository_service_tuf_api/common_models.py
lukpueh: Ah, thanks, I understand. This pydantic v1 syntax ... see https://docs.pydantic.dev/1.10/blog/pydantic-v2-alpha/#changes-to-config:

> To specify config on a model, it is now deprecated to create a class called Config in the namespace of the parent BaseModel subclass. Instead, you just need to set a class attribute called model_config to be a dict with the key/value pairs you want to be used as the config.





Maybe (as a low prio task) we could consider migrating to v2: https://docs.pydantic.dev/2.0/migration/

lukpueh (commented):
Please fix comment, otherwise this looks good to me.

kairoaraujo:
> Please fix comment, otherwise this looks good to me.

I'm filing an issue to update it. ;)

repository_service_tuf_api/common_models.py
kairoaraujo: Yes, I agree
- https://github.com/repository-service-tuf/repository-service-tuf-api/issues/532

lukpueh (approved):

PR #569, opened 2024-03-25
Replace all mentions of "targets" with "artifacts"

MVrachev:
# Description

We want to use "targets" only in three cases:
- the top-level metadata role
- custom target delegation or targets bin role(s)
- when we mention "target files"

This is necessary as currently, we make mistakes in our code which results in messages to the user referring to "artifacts" as "targets" and the opposite. We even had some times where this happened in the same API call.

Signed-off-by: Martin Vrachev [martin.vrachev@broadcom.com](mailto:martin.vrachev@broadcom.com)

# Additional requirements

# Code of Conduct


kairoaraujo (approved):

kairoaraujo:
Merging as breaking changes

PR #574, opened 2024-04-05
Force UTC timezone for all datetime objects

MVrachev:
# Description

The specification defines `datetime` as UTC time zone:
https://theupdateframework.github.io/specification/latest/#file-formats-date-time

The python-tuf project updated all their datetime objects,  so it's time we do it now as well.

I also used the opportunity to add `MOCK_PATH` constants to simplify mocking and 
`fake_datetime` fixture, so we can reuse it.

# Code of Conduct

### DIFF
--- a/repository_service_tuf_api/artifacts.py
+++ b/repository_service_tuf_api/artifacts.py
@@ -4,7 +4,7 @@
 # SPDX-License-Identifier: MIT
 
 import json
-from datetime import datetime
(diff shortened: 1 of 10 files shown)


kairoaraujo (approved):

kairoaraujo (approved):
FT will fail due breaking compatilibity -- FT tested manually.

Issue #647, opened 2024-07-15
Bug: API returns 200 on error


lukpueh:
### What happened?

There are several cases, where the RSTUF API raises an `HTTPException` with status code 200, i.e. returns 2xx on error. This does not seem right wrt HTTP semantics, and also goes against the [FastAPI recommendations](https://fastapi.tiangolo.com/tutorial/handling-errors/).

cc @simi

### What steps did you take?

```bash
$ grep -rnA1 "raise HTTPException(" | grep -B1 200

./repository_service_tuf_api/bootstrap.py:200:        raise HTTPException(
./repository_service_tuf_api/bootstrap.py-201-            status_code=status.HTTP_200_OK,
--
./repository_service_tuf_api/metadata.py:61:        raise HTTPException(
./repository_service_tuf_api/metadata.py-62-            status.HTTP_200_OK,
--
./repository_service_tuf_api/metadata.py:121:        raise HTTPException(
./repository_service_tuf_api/metadata.py-122-            status.HTTP_200_OK,
--
./repository_service_tuf_api/metadata.py:136:        raise HTTPException(
./repository_service_tuf_api/metadata.py-137-            status.HTTP_200_OK,
--
./repository_service_tuf_api/metadata.py:154:            raise HTTPException(
./repository_service_tuf_api/metadata.py-155-                status.HTTP_200_OK,
--
./repository_service_tuf_api/metadata.py:166:            raise HTTPException(
./repository_service_tuf_api/metadata.py-167-                status.HTTP_200_OK,
--
./repository_service_tuf_api/metadata.py:229:        raise HTTPException(
./repository_service_tuf_api/metadata.py-230-            status.HTTP_200_OK,
--
./repository_service_tuf_api/metadata.py:321:        raise HTTPException(
./repository_service_tuf_api/metadata.py-322-            status.HTTP_200_OK,
--
./repository_service_tuf_api/metadata.py:379:        raise HTTPException(
./repository_service_tuf_api/metadata.py-380-            status.HTTP_200_OK,
--
./repository_service_tuf_api/artifacts.py:161:        raise HTTPException(
./repository_service_tuf_api/artifacts.py-162-            status.HTTP_200_OK,
--
./repository_service_tuf_api/artifacts.py:219:        raise HTTPException(
./repository_service_tuf_api/artifacts.py-220-            status.HTTP_200_OK,
```

### What behavior did you expect?

A status code from the 4xx range.

### Code of Conduct

[...]

PR #651, opened 2024-07-19
fix: status code from OK to NOT_FOUND

kairoaraujo:
# Description
This PR fixes the status code from OK (200) to NOT_FOUND (404) when the resource is not available or not in expected state when requested.

Fixes: #647

# Types of changes
- [x] Breaking change (fix or feature that would cause existing functionality not to work as expected)

# Additional requirements

# Code of Conduct

### DIFF
--- a/repository_service_tuf_api/metadata.py
+++ b/repository_service_tuf_api/metadata.py
@@ -59,7 +59,7 @@ def post_metadata(payload: MetadataPostPayload) -> MetadataPostResponse:
     bs_state = bootstrap_state()
     if bs_state.bootstrap is False:
         raise HTTPException(
-            status.HTTP_200_OK,
+            status.HTTP_404_NOT_FOUND,
             detail={
                 "message": "Task not accepted.",
                 "error": (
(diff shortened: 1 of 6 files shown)


repository_service_tuf_api/metadata.py
```diff
@@ -227,7 +227,7 @@ def get_metadata_sign() -> MetadataSignGetResponse:
     # Adds support only when bootstrap is signing state
     if bs_state.bootstrap is False and bs_state.state != "signing":
         raise HTTPException(
-            status.HTTP_200_OK,
+            status.HTTP_404_NOT_FOUND,
```
kairoaraujo, line 230: For this `get metadata sign`, I don't know if we should return 404.
We should keep 200 for this get as the resource is listed in the body.
WDYT? (cc @simi)

lukpueh: > For this get metadata sign, I don't know if we should return 404.

IMO this is the same as the rest. That is, in pre-bootstrap state the resource simply does not exist. So you should return 404, also for consistency. Post-bootstrap , OTOH, if the resource does exist but is empty, it makes sense to return 200 and `{"metadata": {}}`.

For me the important thing is that I don't have to check the [status AND message](https://github.com/pypi/warehouse/blob/6dfb07dd44716fea35cb4ee8e23ee5011e0a2d73/warehouse/tuf/__init__.py#L47-L53) to know if there is an issue.



> We should keep 200 for this get as the resource is listed in the body.

I don't understand.

lukpueh (approved):
Looks good! Thanks for the quick patch.

MVrachev (approved):

PR #966, opened 2026-07-05
build: pin Dockerfile base image and monitor docker images with dependabot

kairoaraujo:
Pins the `python:3.13-slim` base image by digest and adds a `docker` package-ecosystem to dependabot, so base image updates (including future Python version bumps) arrive as dependabot PRs.

The original Python ~~3.14~~ 3.13 bump was dropped from this PR: functional tests cannot pass on3.14 yet because the RSTUF CLI depends on PyKCS11, which has no cp314 wheels until its next release (LudovicRousseau/PyKCS11#155) and its sdist build produces a broken extension on `python:3.14-slim` (`undefined symbol: _ZTVN10__cxxabiv117__class_type_infoE`). The umbrella Pipfile also pins Python 3.13, so a 3.14 move needs to be coordinated across CLI/worker/umbrella once PyKCS11 supports it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -28,3 +28,17 @@ updates:
       github-actions:
         patterns:
           - "*"
+  # Monitor Docker base images
+  - package-ecosystem: "docker"
+    directory: "/"
+    schedule:
+      interval: "daily"
+      time: "10:00"
+    commit-message:
+      prefix: "build"
+      include: "scope"
+    open-pull-requests-limit: 10
+    groups:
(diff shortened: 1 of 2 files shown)


srinjoydutta03 (approved):

kairoaraujo:
Note: rolling back to 3.13 as 3.14 still having issues with PyKCS11 support

srinjoydutta03 (approved):

